PR #61, opened 2022-09-13
UI testing

julian-CStack:


julian-CStack:
cn v2 currencies call may not work as expected so closing this

PR #135, opened 2022-10-14
Update setup.sh and prebuild.sh to add SimpleSwap API key const

sneurlax:
Missing `kSimpleSwapApiKey` in `lib/external_api_keys.dart` throws errors on build, so this addition to setup.sh and prebuild.sh adds an empty default const


sneurlax:
Should PR from origin/, not sneurlax/

PR #150, opened 2022-10-18
Change default bitcoincash testnet electrum node to stackwallet.

danrmiller:
Change default bitcoincash testnet electrum node to use stackwallet's.


danrmiller:
When source branch is external, CI is not working with secrets, so for now I will re-push branch to target repo.

Issue #99, opened 2022-09-29
Cannot import standard wownero 25 word seed phrases


Cypher-Stack:
The only option is for 14 word seed phrases for importing wownero keys.

mvozzo:
> The only option is for 14 word seed phrases for importing wownero keys.
 
Wownero is a fork of Monero meaning the code is nearly identical. 
Stack Wallet currently supports Monero 25 word seed phrase imports, hence, this may be a relatively quick update. 
This feature will allow the Wownero community to migrate Desktop Wallets and CLI Wallets into Stack Wallet.  
Please add this request to the roadmap. Many thanks.

sneurlax:
Closed by #208

PR #131, opened 2022-10-14
Add mainnet Monero and Wownero tests for address generation from test mnemonic

sneurlax:
TODO: Clean up the test wallet folder after the tests conclude
TODO: Test restoring from keys
TODO: Test stack_wallet monero coin service(s) methods
TODO: Test flutter_libmonero methods
~~TODO: Move test mnemonic information to .gitignored file similar to other wallet files~~ :heavy_check_mark:

PR #208, opened 2022-11-07
Enable 25 word Wownero seed imports

sneurlax:
This pull request adds the underlying changes necessary to support 25 word Wownero seed ~~creation and~~ restoration.

 - [ ] ~~Add toggle to Wownero wallet creation page to switch between 14 and 25 word seed lists~~
 - [x] ~~Pass an `int seedWords` of 14 (default) or 25 to the wallet creation function~~
  Not supporting 25 word seed creation yet, although I'd really like to.  It should be as simple as copying Julian's 14-or-25-word toggle on the recovery page to the creation page.
 - [x] Add toggle to Wownero recovery creation page to switch between 14 and 25 word seed lists
 - [x] Copy the Monero height estimation function to Wownero and use it on recover
 - [x] Complete the Wownero height estimation statistics (list the first block every month over the course of Wownero's history)
 - [x] Fix 25 word seed wordlist on recovery

On the backend (in flutter_libmonero), wallet creation and recovery functions have been duplicated and delineated between 14-word and 25-word flavors, with seed checking and validation handled at a higher level in Dart (although flutter_libmonero itself has lower-level checks which will cause calls with ill-formed seeds to fail).  For 14 word seeds, Wowlet's `wownero-seed` (wow_seed) handles seed generation, validation, and recovery, and the official Wownero library handles 25 word seeds.  

See https://github.com/cypherstack/stack_wallet/issues/99


sneurlax:
This supercedes/includes #131, which makes small changes to enable testing of Monero and Wownero and the very most basic library tests.

Issue #212, opened 2022-11-10
Invalid XMR or WOW address not caught by validation


danrmiller:
Version 1.5.16 

On the "Send" screen , changing the capitalization of a Monero or Wownero destination address to invalidate the checksum does not produce an "Invalid Address" error like it does for the other coins.

rehrar:
Fixed in v1.5.17

PR #217, opened 2022-11-10
Use native address validation for cryptonote coins

sneurlax:
Exposes `validate_address` and uses `addressValid` to determine if addresses are valid.  Adds unit tests to Monero and Wownero testing basic address types (*eg.* empty, valid, invalid, wrong nettype, wrong prefix)


sneurlax:
Closes #212

PR #267, opened 2022-12-15
Fix refresh() for Particl when seed has blinded, private, and/or staking tx outputs

sneurlax:
When refreshing a Particl wallet with blinded, private, and/or staking tx outputs, there would be no `['scriptPubKey']['address']` property on certain outputs (blinded, private, and staking tx contain no `['scriptPubKey']` property); and Particl returns addresses as a list `['scriptPubKey']['addresses']<List>`, whereas Stack Wallet uses the more-up-to-date (with respect to Bitcoin core) `['scriptPubKey']['address']<String>`.  This change detects both cases and fixes transaction fetching.


sneurlax:
Ah sorry to be running these unneeded actions, in the future I won't draft PRs so preemptively

PR #268, opened 2022-12-16
Remove Stack Wallet from list of apps with which to open file on Android

sneurlax:
except for `.swbs`


sneurlax:
Please squash :pray:

PR #271, opened 2022-12-22
Add python packages to build instructions and change formatting a little

danrmiller:
Add python packages to build instructions.


AaronFeickert (commented):
Does this need to go into the automated build script?

sneurlax:
> Does this need to go into the automated build script?

I'd think so, yes

PR #319, opened 2023-01-25
Add sandboxed Simplex buy support

sneurlax:
To take from sandbox to production, URLs in SimplexAPI will need to be updated from `simplex-sandbox.stackwallet.com` to `simplex-api.stackwallet.com`, for example


lib/hive/db.dart
```diff
@@ -33,6 +33,7 @@ class DB {
   static const String boxNameDBInfo = "dbInfo";
   static const String boxNameTheme = "theme";
   static const String boxNameDesktopData = "desktopData";
+  static const String boxNameBuys = "buysBox";
```
sneurlax, line 36: could comment this out as it's not used

lib/models/buy/buy_form_state.dart
```diff
@@ -0,0 +1,25 @@
...
+  Buy? _buy;
+  Buy? get buy => _buy;
+  set buy(Buy? value) {
+    _buy = value;
+    // _onBuyTypeChanged();
+  }
+
+  bool reversed = false;
+
+  Future<void> updateEstimate({
```
sneurlax, line 15: Not needed

lib/models/buy/response_objects/pair.dart
```diff
@@ -0,0 +1,73 @@
+import 'dart:ui';
```
sneurlax, line 1: whole file not needed

rehrar (approved):

PR #357, opened 2023-02-21
Epicbox failover - When epicbox server is down default to EU server

Likho:


sneurlax (approved):
Approved and added changes

I changed things around a bit to make what's going on a bit more clear (to me, anyways).  I removed a pair of secureStore reads and writes and returned values directly instead.  I also removed the write to update epic box config in secureStore with the failover as there's no way to reset this setting (yet), nor a toggle between Europe<->Americas if one or the other fails

sneurlax (approved):
I didn't mean to push that comment update and restart tests, sorry, https://github.com/cypherstack/stack_wallet/actions/runs/4237086532/jobs/7362641374 is still valid as 6a6b376 was only a comment update

PR #381, opened 2023-03-02
Fix crash caused by epic wallet deletion panic 

sneurlax:
This switches flutter_libepiccash to a commit on `#fix/delete_wallet` instead of PRing those changes to main.  We should preferably do that differently, or else if that's how we're going to do that, PR that to `flutter_libepiccash#main` then push a pub ref update commit to this PR

### DIFF
--- a/lib/services/coins/epiccash/epiccash_wallet.dart
+++ b/lib/services/coins/epiccash/epiccash_wallet.dart
@@ -227,52 +227,26 @@ Future<String> deleteEpicWallet({
   required String walletId,
   required SecureStorageInterface secureStore,
 }) async {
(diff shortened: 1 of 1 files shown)


sneurlax:
Working on error:
```
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Error("missing field `wallet_dir`", line: 1, column: 117)', src/lib.rs:805:62
```
See https://github.com/cypherstack/flutter_libepiccash/commit/fc4cd4efc6cb84f494578a7697302f107361a8d5

sneurlax:
@Likho's fix tested working, thank you Likho!

PR #434, opened 2023-03-24
Update bitcoin block explorers to use mempool.space

InnocuousFinch:
[mempool.space](https://mempool.space/) is a highly informative bitcoin block explorer with advanced features and a very clean UI. This PR swaps out the chain.so block explorer with mempool.space for both mainnet and testnet bitcoin.

### DIFF
--- a/lib/utilities/block_explorers.dart
+++ b/lib/utilities/block_explorers.dart
@@ -6,13 +6,13 @@ Uri getBlockExplorerTransactionUrlFor({
 }) {
   switch (coin) {
     case Coin.bitcoin:
-      return Uri.parse("https://chain.so/tx/BTC/$txid");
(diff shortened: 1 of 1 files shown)


sneurlax (approved):
The two changes just change two lines changing the block explorer URLs

It's a safe addition, looks good to me

I'd love a way to set the block explorer from within the app ... Or disable it

InnocuousFinch:
Seeing a few people on Twitter asking for a custom block explorer option as well. Definitely agree that users should be able to specify a block explorer of their choice, which gives users the ability to open up transactions in their own self-hosted block explorer.

I'll write up a Feature Request for this, and include providing an option to disable it entirely.

rehrar (approved):

ChuckDoItDoIT (approved):

PR #550, opened 2023-05-24
nano sends

fossephate:
This gets nano sends working, though only just barely
Untested on mobile as I haven't been able to get that working yet


fossephate:
closed - wrong branch

PR #556, opened 2023-05-26
Custom pin length + nmc tx siging fix

julian-CStack:


julian-CStack:
Looks good to me but since this is pin stuff and could lock someone out of their wallet it would be good if another person reviewed it as well

PR #557, opened 2023-05-26
add copyright generating script

detherminal:
The script generates or replaces the current header with the header variable in script file. To update it just change the header variable and run the script again.


julian-CStack:
Android studio's autosave will remove the empty line at the top of a file so it would be good if the script could be modified so as to not add that.

detherminal:
> Android studio's autosave will remove the empty line at the top of a file so it would be good if the script could be modified so as to not add that.

Corrected script for removing space at top

Also added openalias support for xmr :D

PR #626, opened 2023-07-26
Firo clean up and fixes

julian-CStack:


lib/services/coins/firo/firo_wallet.dart
```diff
@@ -4726,7 +4577,13 @@ class FiroWallet extends CoinServiceAPI
...
     final _mnemonic = await mnemonicString;
     final _mnemonicPassphrase = await mnemonicPassphrase;
-    final index = firoGetMintIndex();
+    final lastUsedIndex = await db.isar.lelantusCoins
+        .where()
+        .walletIdEqualTo(walletId)
+        .sortByMintIndexDesc()
+        .mintIndexProperty()
+        .findFirst();
+    final nextFreeMintIndex = (lastUsedIndex ?? 0) + 1;
```
sneurlax, line 4586: this is the only thing that scares me lol

sneurlax: I mean it looks correct, I just worry about lastUsedIndex being unintentionally null, but this worry is based on underdeveloped review--I'm not really sure that lastUsedIndex can be unintentionally null such that we reuse mint index 1

sneurlax (approved):
LGTM

lib/services/coins/firo/firo_wallet.dart
julian-CStack: That index is fetched 3 times. For one of the 3 times it does get double checked against the result of get anonymity set (creating a normal mint). The other 2 places should be doubled checked as well

julian-CStack: I wrapped that specific query in a function so its a) easier to use, and b) mockable https://github.com/cypherstack/stack_wallet/pull/626/commits/6733a367e3461ab40a14ac0a6fb53b22de65701b

Issue #625, opened 2023-07-25
Bug: Shorter mnemonic words selected by autofill when typing in longer word (eg. typing carry, car autofilled, typing actor, act autofilled)


sneurlax:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

On Linux desktop, when you restore a mnemonic phrase which contains words that contain valid words as substrings, the shorter substring is autofilled.

### Expected Behavior

If an input is valid but also itself a substring of a longer valid word, focus should not switch automatically.

### Reproduce Steps

Start typing "actor" and "act" will be recognized as a valid word, with focus switching to the next words' input.  Type "carry", but "car" will be recognized and focus switched.

### Environment

- Operating system and version: Ubuntu 20.04

julian-CStack:
what if `car` is the actual word. Then you do want the focus to switch to next field. This unfortunately a side effect of doing auto-go-to-next-field. Not switching to next field on `car` is not a valid solution as that breaks consistency.
@rehrar thoughts?

detherminal:
I have opened the pr that makes it like: It checks if it is a valid word and there is a longer word available, if both, then it just shows empty (not invalid or valid) and user can go to next word by pressing it (then it shows valid if the previous one is valid). If the user continues writing from car to carry, it normally autocompletes and goes to the next one. I think this is the best practice.

sneurlax:
As referenced [here](https://github.com/cypherstack/stack_wallet/pull/645#issuecomment-1714760700), this issue has been resolved.

PR #630, opened 2023-07-27
fix autocomplete bug

detherminal:
solves #625


sneurlax:
I just need to fix the tests and this will be merged in from origin/detherminal/autocomplete

julian-CStack:
reverted original auto focus

PR #649, opened 2023-08-22
Allow seed words length selection and mnemonic passphrase for new wallets on select coins

julian-CStack:
Adds option screen on new wallet creation for select coins to allow choice of 12 or 24 word seeds as well as an optional mnemonic passphrase. If option is not set to default a dialog will pop up on seed word confirmation game that requests user enters their selected passphrase to verify it before finalizing wallet creation.


AaronFeickert (commented):
No obvious cryptographic issues, but the intent and usage of the passphrase should be made clear to the user.

PR #653, opened 2023-08-24
feat: add tezos api class

detherminal:


julian-CStack (changes requested):
This should be able to be used outside stack wallet. Custom return classes will need to be written for each api call and each api call's function should allow passing in parameters that each respective api endpoint has.

This will mean mean parsing the json returned from the api calls into the newly created classes. 

Then in stack wallet, the stack wallet specific models will need to be constructed using data from the models that the tezos api class returns.

In terms of "just getting it working" none of this is required. But for maintainability and in terms of testing (being able to create mocks) this modular approach is much better

PR #683, opened 2023-10-21
Change rescan logic to scan over maxNumberOfIndexesToCheck, up to highestIndexWithActivity + maxUnusedAddressGap

sneurlax:
Instead of stopping scanning address at `maxNumberOfIndexesToCheck`, we now continue beyond `maxNumberOfIndexesToCheck` up to `highestIndexWithActivity` + `maxUnusedAddressGap`.

Before, if we had activity at address index 1001, this would not be scanned.  Now, we will scan up to 1021 in that case.

`maxNumberOfIndexesToCheck` has been refactored to `minNumberOfIndexesToCheck` in order to better describe what's taking place with this variable now.

`minNumberOfIndexesToCheck` could be reduced to 0: we could just scan up to the gap limit `maxUnusedAddressGap` on each derivation path.  This PR has reduced `minNumberOfIndexesToCheck` to 100, so we scan 5 times the default gap limit of 20, but it could now safely be reduced to 0 or simply removed entirely, only using `highestIndexWithActivity` (or `returningIndex`, in some contexts) plus the gap limit `maxUnusedAddressGap`.

This PR is compatible with the wallet refactor effort ongoing in the `#wallets_refactor` branch.  I feared it would cause merge conflicts, but no!  [stack_wallet/compare/wallets_refactor...rescan](https://github.com/cypherstack/stack_wallet/compare/wallets_refactor...rescan) is clean with no merge conflicts.


sneurlax:
Closes 684

sneurlax:
This functionality is included in https://github.com/cypherstack/stack_wallet/pull/685, but only for Bitcoin Cash

PR #740, opened 2024-01-31
Make settings menu, advanced settings view, and DebugInfoDialog scrollable (if needed) and capitalize certain Firo terms

sneurlax:
- Make entire DebugInfoDialog scrollable so on Windows etc. you can access logs if they overflow the view.
 - Make desktop settings menu scrollable
 - Make advanced settings view scrollable
 - Capitalize certain Firo terms (spark->Spark, lelantus->Lelantus)

### DIFF
--- a/lib/pages/special/firo_rescan_recovery_error_dialog.dart
+++ b/lib/pages/special/firo_rescan_recovery_error_dialog.dart
@@ -209,7 +209,7 @@ class _FiroRescanRecoveryErrorViewState
             children: [
               if (!Util.isDesktop) const Spacer(),
               Text(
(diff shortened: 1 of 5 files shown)


julian-CStack:
Rather than having a scrollable dialog with scrollable content it would be better to shrink the dialog based on screen size and only have the one scrollable element (the logs themselves in this case)

sneurlax:
> Rather than having a scrollable dialog with scrollable content it would be better to shrink the dialog based on screen size and only have the one scrollable element (the logs themselves in this case)

OK, I did that in 8c7f9b491d30d871dd783d80186edd97b7754602

PR #743, opened 2024-02-03
ElectrumX fixes: Use subscribable ElectrumX client for subscribing to chain height, resolve issue with sorting by a null blockHeight for unconfirmed tx, and if just one response is returned, return it as a single-item list

sneurlax:
- [x] Use one chain height subscription per coin (see [ElectrumxChainHeightService](https://github.com/cypherstack/stack_wallet/pull/743/files#diff-539affccbbf6b1154f8e991affde615088dab696612d691cb9fd46bb02817d41))
 - [x] Reconnect on internet failure or other interruption
 - [x] Check chain height subscription validity (ping?)
 - [x] Connect via Tor if enabled
   - [x] Test Tor connections
   - [x] Test reconnection on Tor preference change
 - [x] Add logging
 - [x] Add error handling
 - [x] Disconnect on close wallet if "sync all wallets" not enabled
 - [ ] Add params to StreamController _controller in [ElectrumXSubscription class](https://github.com/cypherstack/stack_wallet/blob/9b93dc78d202b87941c88c959997d567cba702ef/lib/electrumx_rpc/subscribable_electrumx_client.dart#L20).

Also:
 - Resolves issue where a single-item response was returned but an error thrown because it wasn't a single-item List.
 - Resolves issue with sorting an unconfirmed output with a null blockTime for coin selection purposes.


sneurlax:
Reopen once tasks are done.

lib/electrumx_rpc/electrumx_client.dart
```diff
@@ -391,10 +392,14 @@ class ElectrumXClient {
 
       final List<dynamic> response;
       try {
-        response = jsonRpcResponse.data as List;
+        if (jsonRpcResponse.data is Map) {
```
julian-CStack, line 395: As this function is the batch request, the response should always be a list and never a map. Although there is a possibility that if there happens to be a single request in this batched call then it could be the case that a single response is then returned as a map rather than a map in a list of length 1. It may be worth adding an extra check to see of the `requestStrings` list is of length greater than 1 for a more detailed error message.

lib/wallets/wallet/wallet_mixin_interfaces/electrumx_interface.dart
```diff
@@ -793,10 +803,78 @@ mixin ElectrumXInterface<T extends Bip39HDCurrency> on Bip39HDWallet<T> {
...
 
   Future<int> fetchChainHeight() async {
+    final Completer<int> completer = Completer<int>();
+
     try {
-      final result = await electrumXClient.getBlockHeadTip();
-      return result["height"] as int;
-    } catch (e) {
+      // Don't set a stream subscription if one already exists.
+      if (ElectrumxChainHeightService.subscriptions[cryptoCurrency.coin] !=
```
julian-CStack, line 810: This does not guarantee that the subscription will not be set again as we don't know for sure whether the _latestHeight was set so this may not return early

lib/wallets/wallet/wallet_mixin_interfaces/electrumx_interface.dart
```diff
@@ -932,7 +1017,15 @@ mixin ElectrumXInterface<T extends Bip39HDCurrency> on Bip39HDWallet<T> {
 
       // check and add appropriate addresses
       for (int k = 0; k < txCountBatchSize; k++) {
-        int count = counts["${_id}_$k"]!;
+        if (counts["${_id}_$k"] == null) {
```
julian-CStack, line 1020: clean up debug prints

julian-CStack (commented):
Couple small things, looks good otherwise from what I can tell

lib/wallets/wallet/wallet_mixin_interfaces/electrumx_interface.dart
sneurlax: 3d42967c8b5f1a1ae515956b8b80ee87f3f7a444 should address this issue, setting the subscription via null assignment

sneurlax: 46285d44ea0fe1ce8ed83f2930b523e874b3f7d9 should address this

lib/electrumx_rpc/electrumx_client.dart
sneurlax: more logging added in dc6d569433e80bf034de21a46b1bf08219981695

PR #746, opened 2024-02-06
Fix error with Tezos restore

Likho:
Fix error with Tezos wallet restore


sneurlax (approved):

julian-CStack (commented):
What exactly was the issue and how does changing to an explicit synchronous db transaction from an awaited asynchronous db transaction change anything? 
The execution flow should be exactly the same in either case so I am curious as to the reason this specific change works to fix whatever the issue was.

PR #750, opened 2024-02-08
Fulcrum fixes: allow null and empty List responses, reconnect socket if needed, and related changes

sneurlax:
Mostly affects Bitcoin Cash, which uses C++-based Fulcrum servers and may provide different responses than traditional Python-based ElectrumX servers.  Tested to not affect other coins--it seems to add benefit across the board, though.

The `seed blossom` testing mnemonic still presents issues.  Firo wallets can also still show unrelated errors.


sneurlax:
@Likho reported an issue with sending BCH using this branch that's not present on staging, so I converted this to a draft and will double-check sending etc.

PR #751, opened 2024-02-08
Fulcrum: Allow certain null responses, handle unexpected Map and List responses

sneurlax:
Fulcrum, which is used by Litecoin, eCash, and Bitcoin Cash, can provide null responses which we currently treat as an error and for which we throw an error.  This PR allows certain null responses which were found to occur in testing and handles unexpected Map and List responses by wrapping or unwrapping them, as necessary.

See https://github.com/cculianu/Fulcrum/blob/d4b3fa1865fcd2d2bd93d36edb264a719282da98/src/Servers.cpp#L1987C1-L2013C2 for a Fulcrum endpoint which can return null without throwing an error.

Likho had reported an issue with making BCH txs which was introduced by these changes, so I separated all Fulcrum-specific changes from #750 into this PR and put the more innocuous (and generalized) ElectrumX changes into #752.  I was not able to reproduce the issue Likho saw, though, so this PR remains open, as it fixes restoration of Bitcoin Cash and eCash wallets with extensive Fusion history for me.

### DIFF
--- a/lib/electrumx_rpc/electrumx_client.dart
+++ b/lib/electrumx_rpc/electrumx_client.dart
@@ -755,6 +755,13 @@ class ElectrumXClient {
         return {"rawtx": response["result"] as String};
       }
 
+      if (response is List && response.length == 1) {
+        response = response[0];
+        Logging.instance.log(
+            "getTransaction($txHash) returned a single-item list.",
+            level: LogLevel.Warning);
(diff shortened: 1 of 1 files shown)


sneurlax:
Superceded by #760, but I still think handling certain null responses as valid is warranted in certain situations.

PR #754, opened 2024-02-10
ElectrumX result validation

sneurlax:
This adds ElectrumX result validation, which for example will retry a JsonRPC.request based on `bool allowNullOrEmptyResult`, `Type expectedResultType`, and `List<String> expectedResultKeys`.

This is a draft because while these do help force a retry and ultimately provide the correct response, I have been finding that retries lead to a deluge of `_dataHandler found a null req!` messages and then subsequent responses arrive out of order.


sneurlax:
Superceded by #760, however this might still be a good idea.  I like it!

PR #759, opened 2024-02-14
Resolve zero chain height bug

sneurlax:
Thanks @Likho for these changes, I just made a change to check if there's already a completer for the coin being used.  If there is, we await/return it.  Tested working using three Litecoin wallets


sneurlax:
Superceded by #760

Issue #763, opened 2024-02-20
Fix fee estimation responses and remove temporary DOGE hackfix


sneurlax:
Fee estimation especially for Dogecoin (but probably also other low-value coins) is broken.  [This hackfix](https://github.com/cypherstack/stack_wallet/pull/767/files#diff-59dbf188da5b6547df664ab4e558f951f69792b372f40a96432ea1b9103d5b3e) was added as a temporary measure: 

```
        // If the response is -1 or null, return a temporary hardcoded value for
        // Dogecoin.  This is a temporary fix until the fee estimation is fixed.
        if (coin == Coin.dogecoin &&
            (response == null ||
                response == -1 ||
                Decimal.parse(response.toString()) == Decimal.parse("-1"))) {
          // Return 0.05 for slow, 0.2 for average, and 1 for fast txs.
          // These numbers produce tx fees in line with txs in the wild on
          // https://dogechain.info/
          return Decimal.parse((1 / blocks).toString());
          // TODO [prio=med]: Fix fee estimation.
        }
```

As deep as I've delved into the stack (_handleResponse in json_rpc_rpc-3.0.2), the responses all show as -1 or null.

PR #780, opened 2024-02-27
Dogecoin fee estimation hackfix

sneurlax:
See: #763


sneurlax:
This could be improved by also handling failures which lead to large grey boxes on the mobile send view when fees fail to calculate for any reason.  Just force them to fail by returning eg. -1 from estimateFee to reproduce

PR #814, opened 2024-03-18
Fixed build issues on windows

MrCyjaneK:
In addition
- Build machine runs win11 on bare metal
- WSL2 ubuntu 20.04 is used
- https://github.com/miketlk/secp256k1-dll <- flutter build will complain about this one.


lib/wallets/wallet/wallet_mixin_interfaces/cw_based_interface.dart
```diff
@@ -218,9 +218,10 @@ mixin CwBasedInterface<T extends CryptonoteCurrency> on CryptonoteWallet<T>
   }
 
   Future<void> _refreshTxData() async {
+    // Calling the function too quickly causes crashes.
+    await Future<void>.delayed(const Duration(seconds: 1));
```
MrCyjaneK, line 222: This was an experiment, and crashes still occured, less frequently - but this may have nothing to do with the root cause.

PR #824, opened 2024-03-25
bridge sol rpc over localhost and tor

detherminal:
should use stack wallet's official node before releasing to public to list all txs correctly, also the wallet needs optimizations because it is slow


sneurlax:
We don't have an official SOL node yet, not sure if we will set one up...

sneurlax (approved):
The code looks good.  I have not tested it.  Merging should be safe as we would still need to test in `solana` branch and merge to `staging`.

PR #841, opened 2024-04-25
Remove unused socks_socket package

sneurlax:
So the nonexistence of github.com:cypherstack/socks_socket doesn't interfere with `pub get`s

We don't use socks_socket.dart from this package, instead using it via cypherstack/tor.

Reference https://github.com/cypherstack/tor/pull/25

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -1574,7 +1574,7 @@ packages:
     source: hosted
     version: "1.0.4"
   socks_socket:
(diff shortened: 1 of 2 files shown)


sneurlax:
Then cypherstack/socks_socket should be deleted.

sneurlax:
Superseded by https://github.com/cypherstack/stack_wallet/pull/844

PR #845, opened 2024-04-29
fix deprecated fee func and null fees.

detherminal:
Replaces the deprecated Solana fee function `getFees()` with `getFeeForMessage()`, also adds exception for null fees. tested and works.


sneurlax (approved):
LGTM!  I just wonder if we always ought to ask for the fees for 1 SOL--or how far off the estimate would be for everyday send amounts.  The difference shouldn't be drastic so a 1 SOL tx should be a decent fee approximation.

PR #846, opened 2024-04-29
solana fee estimation fixes

detherminal:
:)

### DIFF
--- a/lib/wallets/wallet/impl/solana_wallet.dart
+++ b/lib/wallets/wallet/impl/solana_wallet.dart
@@ -64,8 +64,7 @@ class SolanaWallet extends Bip39Wallet<Solana> {
         fundingAccount: pubKey,
         recipientAccount: pubKey,
         lamports: transferAmount.raw.toInt(),
-      ),
-      ComputeBudgetInstruction.setComputeUnitPrice(microLamports: 6000),
+      )
     ]).compile(recentBlockhash: latestBlockhash!.value.blockhash, feePayer: (await _getKeyPair()).publicKey);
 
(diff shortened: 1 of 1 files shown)


lib/wallets/wallet/impl/solana_wallet.dart
```diff
@@ -66,10 +65,8 @@ class SolanaWallet extends Bip39Wallet<Solana> {
         recipientAccount: pubKey,
         lamports: transferAmount.raw.toInt(),
       ),
-    ]).compile(
-      recentBlockhash: latestBlockhash!.value.blockhash,
-      feePayer: pubKey,
-    );
+      ComputeBudgetInstruction.setComputeUnitPrice(microLamports: 6000),
```
sneurlax, line 68: Why 6000 here?

detherminal: transfer instruction is 5000 lamports and compute unit budget instruction is 1000 lamports, total of 6k lamports

julian-CStack: > transfer instruction is 5000 lamports and compute unit budget instruction is 1000 lamports, estimated total of 6k lamports

Include this explanation in a comment please so the seemingly random 6000 is documented in code

detherminal:
Current approach of fee:
- Get only necessary transfer fee
- Add fee on top of base if it is higher than it
- Limit the tx to not use more than the specified amount

Also fixed the sent to self bug, it is working correctly now

sneurlax (approved):
Looks good to me!

PR #838, opened 2024-04-24
monero.dart clean implementation

MrCyjaneK:
depends on https://github.com/cypherstack/flutter_libmonero/pull/64

build instructions:

```
$ git clone https://github.com/MrCyjaneK/stack_wallet --branch cyjan-monerodartv2 --recursive
$ cd stack_wallet
$ rm -rf crypto_plugins/flutter_libmonero
$ git clone https://github.com/mrcyjanex/flutter_libmonero --branch cyjan-monerodartnew crypto_plugins/flutter_libmonero
$ cd scripts && ./prebuild.sh && cd ..
$ cd scripts/linux && ./build_all.sh && cd ../..
$ cd scripts/android && ./build_all.sh && cd ../..
```

PR #848, opened 2024-05-03
Monerodart -> testing integration

sneurlax:
Reference #838; basically #838 + `testing` branch (which includes Bitcoin Frost and other changes)


android/app/src/main/jniLibs/arm64-v8a/libmonero_libwallet2_api_c.so
julian-CStack, line 1: delete

android/app/src/main/jniLibs/armeabi-v7a/libmonero_libwallet2_api_c.so
julian-CStack, line 1: delete

android/app/src/main/jniLibs/x86_64/libmonero_libwallet2_api_c.so
julian-CStack, line 1: delete

scripts/android/build_all.sh
```diff
@@ -12,11 +12,11 @@ mkdir -p build
 
 PLUGINS_DIR=../../crypto_plugins
 
-(cd "${PLUGINS_DIR}"/flutter_liblelantus/scripts/android && ./build_all.sh ) &
-(cd "${PLUGINS_DIR}"/flutter_libepiccash/scripts/android && ./build_all.sh )  &
-(cd "${PLUGINS_DIR}"/flutter_libmonero/scripts/android/ && ./build_all.sh  ) &&
-set_rust_to_1720 &&
-(cd "${PLUGINS_DIR}"/frostdart/scripts/android && ./build_all.sh ) &
+(cd "${PLUGINS_DIR}"/flutter_liblelantus/scripts/android && ./build_all.sh )
+(cd "${PLUGINS_DIR}"/flutter_libepiccash/scripts/android && ./install_ndk.sh && ./build_openssl.sh && ./build_all.sh )
```
julian-CStack, line 16: libepiccash's build all script runs install ndk and openssl (or at least it did last I checked)

macos/wownero_libwallet2_api_c.dylib
julian-CStack, line 1: delete

macos/monero_libwallet2_api_c.dylib
julian-CStack, line 1: delete

lib/wallets/wallet/impl/wownero_wallet.dart
```diff
@@ -330,7 +337,9 @@ class WowneroWallet extends CryptonoteWallet with CwBasedInterface {
         //     date: DateTime.now().subtract(const Duration(
         //         days:
         //         2))); // subtract a couple days to ensure we have a buffer for SWB
-        final bufferedCreateHeight = getSeedHeightSync(wallet!.seed.trim());
+        // TODO(mrcyjanek): implement
```
julian-CStack, line 340: needs to be done asap

lib/wallets/wallet/impl/wownero_wallet.dart
```diff
@@ -441,7 +450,8 @@ class WowneroWallet extends CryptonoteWallet with CwBasedInterface {
 
         // extract seed height from 14 word seed
         if (seedLength == 14) {
-          height = getSeedHeightSync(mnemonic.trim());
+          // TODO(mrcyjanek): implement
+          height = 1; // getSeedHeightSync(mnemonic.trim());
```
julian-CStack, line 454: needs to be done asap

.fvmrc
julian-CStack, line 1: delete (otherwise we need to update this every time we update the flutter version)

.vscode/settings.json
julian-CStack, line 1: revert this change

.gitignore
julian-CStack, line 1: keeping this won't hurt since we don't need to update this

android/app/src/main/AndroidManifest.xml
julian-CStack, line 1: Any reason extractNativeLibs is specifically set here?
If there is no specific reason then I suggest reverting and letting the default be chosen by the system as described in: https://developer.android.com/guide/topics/manifest/application-element

android/app/build.gradle
julian-CStack, line 1: Reasons for this?

crypto_plugins/flutter_libmonero
julian-CStack, line 1: This requires a separate review of the changes to flutter_libmonero

julian-CStack (changes requested):
see comments

android/app/src/main/AndroidManifest.xml
MrCyjaneK: for the .so to be extracted, the ones that are marked for deletion. I'll make it a cmake install step instead if symlinking is not desired

android/app/build.gradle
MrCyjaneK: monero_c also depends on libc++_shared.so, and in early versions of this PR i was getting build errors because multiple packages were providing the dynamic library

scripts/android/build_all.sh
sneurlax: Yes it does.  Resolved by ce17fed2bdc21147ed21164d505dd01d1feac2fbe4420f8c561ba6cd0da91607

.gitignore
sneurlax: but I use fvm on macos :P

.vscode/settings.json
sneurlax: Resolved by 89a687b9715a8a69590f01da76b83c2a8424c4ee

MrCyjaneK:
@julian-CStack regarding .so/.dylib symlinks deletion, what is the preferred way to include them in the apk/.app on android/macos?

lib/wallets/wallet/impl/monero_wallet.dart
```diff
@@ -218,8 +277,8 @@ class MoneroWallet extends CryptonoteWallet with CwBasedInterface {
           final addressString =
               (CwBasedInterface.cwWalletBase as MoneroWalletBase?)
                   ?.getTransactionAddress(
-            addressInfo!['accountIndex'] as int,
-            addressInfo['addressIndex'] as int,
+            addressInfo?['accountIndex'] as int? ?? 0,
+            addressInfo?['addressIndex'] as int? ?? 0,
```
julian-CStack, line 281: Is it ok to default to 0 here? Or should there be some error?

MrCyjaneK: should be fine, there is no account switching feature in stack yet.

julian-CStack:
merged https://github.com/cypherstack/stack_wallet/tree/monerodart_testing into https://github.com/cypherstack/stack_wallet/tree/monero_merge which is based on current staging that includes a pile of generalization changes

PR #885, opened 2024-06-07
Add defaultFeeRate per ElectrumX coin and use it if fee estimates unavailable from node

sneurlax:
defaultFeeRate as used here may instead more accurately refer to the minimum (absolute) fee for relay.

### DIFF
--- a/lib/electrumx_rpc/electrumx_client.dart
+++ b/lib/electrumx_rpc/electrumx_client.dart
@@ -10,6 +10,7 @@
 
 import 'dart:async';
 import 'dart:io';
+import 'dart:math';
 
 import 'package:connectivity_plus/connectivity_plus.dart';
 import 'package:decimal/decimal.dart';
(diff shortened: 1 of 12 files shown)


lib/wallets/crypto_currency/interfaces/electrumx_currency_interface.dart
julian-CStack, line 1: Better to add the minfee rate param to BitcoinFrost class as well so we don't need to piggyback off Bitcoin

lib/electrumx_rpc/electrumx_client.dart
```diff
@@ -1113,17 +1115,23 @@ class ElectrumXClient {
...
-          // Return 0.05 for slow, 0.2 for average, and 1 for fast txs.
-          // These numbers produce tx fees in line with txs in the wild on
-          // https://dogechain.info/
-          return Decimal.parse((1 / blocks).toString());
-          // TODO [prio=med]: Fix fee estimation.
+        // If the response is -1 or null, fall back to the defaultFeeRate.
+        if (response == null ||
+            response == -1 ||
+            Decimal.parse(response.toString()) == Decimal.parse("-1")) {
+          if (CryptoCurrency is! BitcoinFrost) {
```
julian-CStack, line 1122: `if (CryptoCurrency is! BitcoinFrost)` should be `if (cryptoCurrency is! BitcoinFrost)`

sneurlax:
Those requested changes and the BigInt usage from chat have all been resolved by 713d8b0cde302b1d864052ace61850fb406613be

julian-CStack (approved):

PR #897, opened 2024-06-27
Bump solana to 1.122.1

sneurlax:
*with tor features ported forward**


sneurlax:
converted to draft because most of these changes aren't actually to `espresso-cash-public/solana`, but rather to the greater `espresso-cash-public`.

sneurlax:
There are since new actually-relevant Solana changes which we should track, imo

Issue #684, opened 2023-10-21
Activity at indices over maxNumberOfIndexesToCheck is ignored


sneurlax:
`maxNumberOfIndexesToCheck` is currently set to 1000.  Activity at indices over this limit, eg at index 1001, is ignored.

Instead of a hard cap on wallet activity, this should be a dynamic limit based on the highest index with activity `highestIndexWithActivity` + the gap limit `maxUnusedAddressGap`.  Then, `maxNumberOfIndexesToCheck` would be better described as `minNumberOfIndexesToCheck`, because we will always scan up to either `minNumberOfIndexesToCheck` or `highestIndexWithActivity` + `maxUnusedAddressGap`, whichever is higher.

Then, `minNumberOfIndexesToCheck` could be reduced to 0 or simply removed and we could just scan up to the gap limit `maxUnusedAddressGap` on any derivation paths.  Short or removing `minNumberOfIndexesToCheck` entirely, it could safely be reduced from 1000 to 100 or any other lower number.  I recommend removing `minNumberOfIndexesToCheck` entirely (or setting it to 0) and increasing `maxUnusedAddressGap` from 20 to 42.

sneurlax:
#683 implements the above, except leaving `maxUnusedAddressGap` at 20.

sneurlax:
The fusion testing wallet goes up to index 4164 currently btw

sneurlax:
Closed by #900

PR #900, opened 2024-06-27
Reconfigure logic for rescan to remove upper address limit

sneurlax:
Now we just keep scanning up until the unused activity gap is reached.


sneurlax:
Closes #684

PR #904, opened 2024-07-01
Quick toggle balances as appropriate

sneurlax:
If only 2 balances exist, toggle them on press instead of showing the dialog.

Also removes the chevron/down arrow icon based on this same criteria.


julian-CStack:
Needs work

Issue #832, opened 2024-04-10
[Feature Request] Add a toggle button to disable the address generator


Senophyx:
Add a feature to disable address generation when the address is already in use. Create a toggle button in the Advanced menu for example. I know this is not suitable for privacy issues, but for people like me it might be a bit annoying if I have to constantly look at new addresses. 

Thanks.

PR #910, opened 2024-07-03
Add advanced settings option to disable new address generation by default

sneurlax:
The "Generate new address" button still exists and works.


sneurlax:
Closes #832

julian-CStack:
Needs more work. Currently setting appears in all wallets but is ignored for anything besides an electrumx based coin. Also `checkReceivingAddressForTransactions` should not be called at all if setting enabled

sneurlax:
OK, `checkReceivingAddressForTransactions` isn't called if the pref is set

Except for on recovery, because at that point `reuseAddress` can't even have been set

My only lingering issue with this feature is that I want all addresses to be checked up to the gap limit, but only want the last manually-generated address to be shown.  That would make this feature a little more practical

PR #924, opened 2024-07-06
update stack.svg in campfire light theme

sneurlax:


sneurlax:
doesn't fix svg afaict

PR #941, opened 2024-07-16
Fix cast in case of non-BIP39 coins

sneurlax:
Issue:

Screenshot from 2024-07-16 10-25-16

### DIFF
--- a/lib/pages_desktop_specific/my_stack_view/wallet_view/sub_widgets/desktop_receive.dart
+++ b/lib/pages_desktop_specific/my_stack_view/wallet_view/sub_widgets/desktop_receive.dart
@@ -190,11 +190,16 @@ class _DesktopReceiveState extends ConsumerState<DesktopReceive> {
       if (supportsSpark) {
         _walletAddressTypes.insert(0, AddressType.spark);
       } else {
-        _walletAddressTypes.addAll(
-          (wallet as Bip39HDWallet)
-              .supportedAddressTypes
-              .where((e) => e != wallet.info.mainAddressType),
-        );
+        if (wallet is Bip39HDWallet) {
+          _walletAddressTypes.addAll(
+            wallet.supportedAddressTypes
+                .where((e) => e != wallet.info.mainAddressType),
+          );
+        } else {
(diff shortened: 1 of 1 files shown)


lib/pages_desktop_specific/my_stack_view/wallet_view/sub_widgets/desktop_receive.dart
```diff
@@ -190,11 +190,16 @@ class _DesktopReceiveState extends ConsumerState<DesktopReceive> {
...
-          (wallet as Bip39HDWallet)
-              .supportedAddressTypes
-              .where((e) => e != wallet.info.mainAddressType),
-        );
+        if (wallet is Bip39HDWallet) {
+          _walletAddressTypes.addAll(
+            wallet.supportedAddressTypes
+                .where((e) => e != wallet.info.mainAddressType),
+          );
+        } else {
```
julian-CStack, line 198: This `else` block should be removed as it does nothing and if it is not a `Bip39HDWallet` then we do not have more than the a single address type per wallet that I'm aware of

julian-CStack (changes requested):

lib/pages_desktop_specific/my_stack_view/wallet_view/sub_widgets/desktop_receive.dart
julian-CStack: This issue is caused above at https://github.com/cypherstack/stack_wallet/blob/5eb65ba13e80dd08779c8f5c6b1c5002c0ced5f3/lib/pages_desktop_specific/my_stack_view/wallet_view/sub_widgets/desktop_receive.dart#L184
It should be similar to: https://github.com/cypherstack/stack_wallet/blob/6690581b518e9ad68f8fb501e7a92c533365d041/lib/pages/receive_view/receive_view.dart#L186 I believe

julian-CStack:
Needed this for frost wallets so it been fixed by https://github.com/cypherstack/stack_wallet/pull/943/commits/18b6de93e86894f471451bef2546c84819b2023d

PR #967, opened 2024-08-08
do not prepend https:// for .onion domains on mobile

sneurlax:


sneurlax (commented):
I would like this to be prettier/better/cleaned up, but it works

PR #973, opened 2024-08-28
coin control tweaks

julian-CStack:


lib/pages/send_view/send_view.dart
```diff
@@ -881,6 +881,48 @@ class _SendViewState extends ConsumerState<SendView> {
...
+            utxos.map((e) => BigInt.from(e.value)).reduce((v, e) => v += e),
+        fractionDigits: ref.read(pWalletCoin(walletId)).fractionDigits,
+      );
+
+  Future<void> _sendAllTapped(bool showCoinControl) async {
+    final Amount amount;
+
+    if (showCoinControl && selectedUTXOs.isNotEmpty) {
+      amount = _selectedUtxosAmount(selectedUTXOs);
+    } else if (isFiro) {
```
sneurlax, line 903: My only qualm so far is this else if, I just need to read over it a few more times to grok if this is going to be an issue or not (probably not)

lib/wallets/wallet/wallet_mixin_interfaces/electrumx_interface.dart
```diff
@@ -220,7 +223,8 @@ mixin ElectrumXInterface<T extends ElectrumXCurrencyInterface>
     // gather required signing data
     final utxoSigningData = await fetchBuildTxData(utxoObjectsToUse);
 
-    if (isSendAll) {
+    if (isSendAll || isSendAllCoinControlUtxos) {
+      assert(satoshiAmountToSend == satoshisBeingUsed);
```
sneurlax, line 227: assert in prod?

sneurlax (approved):
LGTM other than the assert in prod (which is fine really)

PR #978, opened 2024-09-05
Defer network calls from init to fetchChainHeight or serverCanBatch

sneurlax:


julian-CStack (approved):
Nothing jumps out at me so if it was thoroughly tested then it should be good to go.

PR #1017, opened 2024-11-14
View only wallet stuff mainly

julian-CStack:
Some bug fixes as well

### DIFF
--- a/lib/wallets/crypto_currency/coins/wownero.dart
+++ b/lib/wallets/crypto_currency/coins/wownero.dart
@@ -92,7 +92,7 @@ class Wownero extends CryptonoteCurrency {
   bool get hasMnemonicPassphraseSupport => false;
 
   @override
-  List<int> get possibleMnemonicLengths => [defaultSeedPhraseLength, 25];
+  List<int> get possibleMnemonicLengths => [defaultSeedPhraseLength, 16, 25];
 
   @override
   BigInt get satsPerCoin => BigInt.from(100000000000);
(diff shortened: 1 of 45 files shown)


lib/wallets/crypto_currency/coins/wownero.dart
```diff
@@ -92,7 +92,7 @@ class Wownero extends CryptonoteCurrency {
   bool get hasMnemonicPassphraseSupport => false;
 
   @override
-  List<int> get possibleMnemonicLengths => [defaultSeedPhraseLength, 25];
+  List<int> get possibleMnemonicLengths => [defaultSeedPhraseLength, 16, 25];
```
sneurlax, line 95: Wownero uses a seed length of 14.  Edit: correction, 14, 16, and 25 are all valid: https://guides.cakewallet.com/docs/cryptos/wownero/#seed-format

sneurlax: Wowlet generates seeds of length 14, Cake generates 16 by default.

sneurlax (changes requested):
This line https://github.com/cypherstack/stack_wallet/pull/1017/files/49b0b7c0bf2a775fef41aebcd1cd6f70067bb751#diff-725f7b92c3c085b20b03035871c1e51326a77feec2517b2d5a8f6347826f8a7cR95 can be changed to `14, 16, 25` for Wownero, but that's ultimately a design/compatibility choice, so this request for changes can also be accepted as an approval if we just don't want to support 14 word seeds (I think they're dumb, but on the other hand Wowlet uses them, so ¯\_(ツ)_/¯)

sneurlax (approved):

Issue #954, opened 2024-07-24
Bug: Add Fastlane documentation to repository.


RokeJulianLockhart:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

https://github.com/cypherstack/stack_wallet/tree/9c37f99e9e74be3fea990aa993a4f09eec403d9e does not contain Fastlane documentation.

### Expected Behavior

 should contain 

### Reproduce Steps

https://gitlab.com/fdroid/rfp/-/jobs/7420581232

### Environment

- Operating system and version: Debian Bookwyrm
- Device platform and version: https://gitlab.com/fdroid/rfp/-/pipelines/1386690875
- Real device or emulator/simulator: GitLab Pipeline

### Further Information

For:

1.	https://fdroid.stackwallet.com/?fingerprint=764B4262F75750A5F620A205CEE2886F18635FBDA18DF40758F5A1A45A950F84 per https://github.com/cypherstack/stack_wallet/issues/72#issuecomment-1263675783
1.	https://gitlab.com/fdroid/rfp/-/issues/2819#note_2014406448
1.	https://github.com/cypherstack/stack_wallet/issues/385#issue-1608501521

sneurlax:
Re: #72, inclusion on F-Droid has been approved

@RokeJulianLockhart what's needed from us for this?  Do you have any examples?

RokeJulianLockhart:
> https://github.com/cypherstack/stack_wallet/issues/954#issuecomment-2256547591

@sneurlax, you might find https://zaneschepke.hashnode.dev/publishing-to-f-droid#heading-fastlane-setup useful.

PR #1020, opened 2024-11-15
Add fastlane metadata

singpolyma:
For #954


singpolyma:
Or maybe fdroid would be ok with this in the android subdirectory? not sure

julian-CStack:
> Or maybe fdroid would be ok with this in the android subdirectory? not sure

Not sure. I'm not very familiar with fdroid or fastlane.

Looking at https://docs.flutter.dev/deployment/cd#fastlane it seems it should be?

singpolyma:
@julian-CStack it's not about what fastlane supports but what fdroid supports. It seems they look in a small number of places so in root like this is probably best. https://gitlab.com/fdroid/rfp/-/issues/2819#note_2213442386

julian-CStack:
> @julian-CStack it's not about what fastlane supports but what fdroid supports. It seems they look in a small number of places so in root like this is probably best. https://gitlab.com/fdroid/rfp/-/issues/2819#note_2213442386

I would expect that if fdroid depends on fastlane, fdroid would work with what fastlane supports. But like I said before, I haven't don't have any experience with either fdroid or fastlane.

I'm ok with whatever works.

danrmiller (dismissed):
LGTM. Hey buddy!

Issue #1061, opened 2024-12-26
Bug: Login/Password of Node dialog can't be made empty after setting a value


Cyrix126:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Exist at least on [v2.1.9](https://github.com/cypherstack/stack_wallet/releases/tag/build_256)

Saving an empty value after setting a non empty value will keep the non empty value.

### Expected Behavior

When putting an empty value after putting a non empty value and saving, the empty value should persist.

### Reproduce Steps

1. Put a value into the fields Login and/or Password 
2. Save
3. Empty the fields
4. Save
5. Exit the dialog
6. Come back to the dialog
7. See that the old non empty value is still present

### Environment

- Artix Linux Desktop

PR #1062, opened 2024-12-26
fix: Login/Password of Node dialog can't be made empty after setting a value

Cyrix126:
Fix https://github.com/cypherstack/stack_wallet/issues/1061

The provider value will not be updated if set to null. So in this PR it will be set to the empty String instead of null.


sneurlax (approved):

sneurlax:
I overlooked that using an empty string has consequences because null is expected for certain cases

julian-CStack (approved):

Issue #1008, opened 2024-10-17
Bug: Unable to connect to remote node


deverickapollo:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Configuration of local node fails with Server Unreachable when pressing the "Test Connection" button. 

### Expected Behavior

Configure URL / Port / Login / Password should result in a successful connection to the remote node.

### Reproduce Steps

1. Install monerod with Umbrel
2. Configure settings for remote node access
3. Test Connection results in failure

### Environment

- Operating system and version: Stack Wallet - latest version with monerod v0.18.3.3
- Device platform and version: monerod is packaged with UmbrelOS. Environment settings can be found in the docker compose file: https://github.com/getumbrel/umbrel-apps/blob/master/monero/docker-compose.yml

### Logs

https://gist.github.com/deverickapollo/920dfe41ab8f984076c5f610acd8b862

### Further Information

To provide additional insight, connectivity between the remote node works for all other wallets including CakeWallet, Monerujo, and Feather Wallet for Desktop. This issue seems to be specific to Stacks Wallet and has been reported by several users.

Node details:
Configured server URL: http://10.0.1.3
Port: 18081
Login: monero
Password: xxxxxx

sneurlax:
Does this use RPC authentication?  If so, it's probably related to https://github.com/cypherstack/stack_wallet/issues/981

deverickapollo:
👍 RPC auth is enabled on the node so this should be the same issue.

Issue #981, opened 2024-09-09
Bug: Monero - server unreachable when rpc-login is ON


mariodian:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When I try connecting to my own monero node, I'm prompted to confirm untrusted certificate. As soon as I click `trust` I get the `server unreachable` error. I'm a 100% sure I provide the correct login credentials since I have no issues connecting to my node in Cake Wallet under these circumstances. 

I can connect to the node with Stack when `rpc_login` is NOT set.

### Expected Behavior

After clicking `trust` I should be able to connect to the server because the app can obviously communicate with it since it shows me the certificate fingerprint. 

### Reproduce Steps

1) `monerod.conf`:
set `rpc-login=<user>:<password>`
2) add new node to Stack with the correct login credentials
3) click "test connection"
4) click "trust" certificate

Error

### Environment

Server:
- Operating system and version: Ubuntu 24.04 LTS
- Device platform and version: Intel NUC 13

Client:
- Operating system and version: iOS 17.6.1
- Device platform and version: iPhone 15 Pro, Stack 2.1.5

sneurlax:
Can you provide an example URL you would use in Cake Wallet that works there?  I haven't tested with RPC login credentials yet...

sneurlax:
Working on this, just still having issues with RPC digest authentication

GhostDog98:
Here's a video demo, as well as a (seemingly related?) Issue where the app insists on adding `http://` to a local IP address; https://youtu.be/T0Xy2B8h7ZQ

mariodian:
> Issue where the app insists on adding `http://` to a local IP address; https://youtu.be/T0Xy2B8h7ZQ

That’s fine. The rpc calls are made over http.

sneurlax:
I made this tool which should now be able to resolve this issue: https://pub.dev/packages/digest_auth

Kreyren:
Reproducible via NiXium's Monero Node: https://github.com/Arcanyx-org/NiXium/tree/central?tab=readme-ov-file#monero-node

Still says server not reachable on version 2.1.9

GhostDog98:
Is there a timeline on a possible fix for this? I'm concerned about using stack wallet now due to the lack of being able to connect to my own node.
As has recently been discovered, a malicious node can undermine the effects of ring-CT, and we have evidence that malicious actors are running such nodes.

GhostDog98:
[...]

PR #1063, opened 2024-12-27
handle authentication in test of monero nodes

Cyrix126:
fix https://github.com/cypherstack/stack_wallet/issues/981
fix https://github.com/cypherstack/stack_wallet/issues/1008


lib/utilities/test_monero_node_connection.dart
```diff
@@ -210,3 +200,18 @@ Future<bool> showBadX509CertificateDialog(
 
   return result ?? false;
 }
+
+extension on SOCKSSocket {
```
sneurlax, line 204: Cool, I should look at adding this to SOCKSSocket itself!

PR #1092, opened 2025-03-12
fix: use trocador provider icons

detherminal:
Use Trocador's providers' icons instead of the default Trocador icon, you may also want to add "(via Trocador)" to the provider names. Fetches the logo from the Trocador API's `provider_logo` key as a url string, fallbacks to default icon if there isn't.

Before:
before

After:
after


b4n6-b4n6:
what about Quickex and stealthEx icons? I see the fallback default icon

detherminal:
> what about Quickex and stealthEx icons? I see the fallback default icon

The link provided by Trocador is giving 404, that's why they fallback. Not an issue on our side, but Trocador's side.

detherminal:
> do I understand correctly that all the logos returned by the trocador API are consistently SVGs? In that case I think yes it's reasonable to just accept it like this (we could also hardcode the logos, but that seems silly/wrong approach)

They are not SVGs but PNGs and the API returns image URLs. Also, hardcoding is not a good idea as providers can change over time and hardcoding creates a technical/practical debt.

julian-CStack (approved):
LGTM

PR #1097, opened 2025-03-17
feat: detect bitcoin/monero uri's

### LINKED ISSUE #1069: Feature request: Parse bitcoin/monero URIs when pasted into My Wallet -> Send Amount or Send Address
As a user of monero and bitcoin when I want to pay for an invoice I like to copy the monero/bitcoin URI
Sadly Stack Wallet does not support parsing of monero/bitcoin URI

How I see it...
This feature would intercept pasting of text into monero/bitcoin send address and send amount text input fields
When detected pasting of bitcoin or monero scheme app would cancel pasting of text followed by programmatically inserting parsed address and/or amount into address and/or amount fields followed by changing of focus to either amount text input field (if still empty) or to send button

I can bounty this 1 xmr

Here pages for monero/bitcoin URI
https://github.com/monero-project/monero/wiki/URI-Formatting
https://en.bitcoin.it/wiki/BIP_0021

detherminal:
Closes #1069

Detects URI Schemes for Bitcoin and Monero and uses them.


julian-CStack (changes requested):
Would be better to handle all URIs for all coins. That is somewhat set up already here: https://github.com/cypherstack/stack_wallet/blob/5adfee831ec008d4e071ea5043c7502e29e7bfc7/lib/utilities/address_utils.dart#L133-L133
That is already used in several places but mostly just for handling QR code reads such as here: https://github.com/cypherstack/stack_wallet/blob/5adfee831ec008d4e071ea5043c7502e29e7bfc7/lib/pages/send_view/send_view.dart#L138-L138
That `_scanQr()` function can probably be refactored into two functions. One for the scanning and a second for handing the parsing of the content. That parsing function can then be used on paste as well.

Rather than attempting a full check and parse every time a character is typed it would be better to hook into the on paste event. See https://api.flutter.dev/flutter/material/TextField/contextMenuBuilder.html and https://api.flutter.dev/flutter/material/AdaptiveTextSelectionToolbar-class.html

Finally, this PR only applies to the standard mobile send screen when it should be consistent across all send forms where it would make sense to have this such as on desktop (and possibly some other places as well. Would need to look around the code to find out).

detherminal:
> Would be better to handle all URIs for all coins. That is somewhat set up already here:
> 
> https://github.com/cypherstack/stack_wallet/blob/5adfee831ec008d4e071ea5043c7502e29e7bfc7/lib/utilities/address_utils.dart#L133-L133
> 
> That is already used in several places but mostly just for handling QR code reads such as here:
> 
> https://github.com/cypherstack/stack_wallet/blob/5adfee831ec008d4e071ea5043c7502e29e7bfc7/lib/pages/send_view/send_view.dart#L138-L138
> 
> That `_scanQr()` function can probably be refactored into two functions. One for the scanning and a second for handing the parsing of the content. That parsing function can then be used on paste as well.
> 
> Rather than attempting a full check and parse every time a character is typed it would be better to hook into the on paste event. See https://api.flutter.dev/flutter/material/TextField/contextMenuBuilder.html and https://api.flutter.dev/flutter/material/AdaptiveTextSelectionToolbar-class.html
> 
> Finally, this PR only applies to the standard mobile send screen when it should be consistent across all send forms where it would make sense to have this such as on desktop (and possibly some other places as well. Would need to look around the code to find out).

I think this `contextMenuBuilder` does not detect keyboard paste events like Ctrl+P which is almost what everybody use. Do you think it should be that way? This is the way that covers almost all use cases.

detherminal:
Seperated the `_scanQr` function into two pieces as well as the ones in `desktop_send`.

PR #1123, opened 2025-04-25
feat(tor): activate tor killswitch independent of tor connection or pref

sneurlax:


sneurlax:
Converted to draft due to staleness.  Can probably close and delete branch safely

PR #1141, opened 2025-05-25
feat: add paper wallet functionality

detherminal:
- Adds functionality to import paper wallets by scanning QRs, in the right top part of the coin selection menu (as discussed with rehrar).
- Adds respective functions to each coin for specific functionality.
- Imports the paper wallets automatically.
- Adds [Monero gift wallet functionality](https://github.com/monero-project/meta/issues/729) by scanning and checking out the balance to a new secure wallet if the parameter `txid` is provided.


lib/wallets/crypto_currency/coins/monero.dart
```diff
@@ -122,6 +144,105 @@ class Monero extends CryptonoteCurrency {
...
+      await (newWallet as MoneroWallet).init(wordCount: 16);
+      await newWallet.info.setMnemonicVerified(isar: ref
+          .read(mainDBProvider)
+          .isar);
+      ref.read(pWallets).addWallet(newWallet);
+
+      await newWallet.open();
+      await newWallet.generateNewReceivingAddress();
+
+      // Send the balance to the new wallet
```
sneurlax, line 208: There's no information dialogs or opportunity for a user to approve or deny the sweep transaction.  See usages of ConfirmTransactionView for how we do so in Stack

lib/wallets/crypto_currency/coins/monero.dart
```diff
@@ -122,6 +144,105 @@ class Monero extends CryptonoteCurrency {
...
+      final txData = await wallet.prepareSend(txData: TxData(
+        recipients: [
+          (address: (await newWallet.getCurrentReceivingAddress())!.value, amount: (await wallet.totalBalance), isChange: false)
+        ],
+        feeRateType: FeeRateType.average,
+      ));
+      await wallet.confirmSend(txData: txData);
```
sneurlax, line 215: What happens to wallet after this send?  This temporary wallet should at least be closed and probably deleted.  That should also require user input similar to the previous recommendation to use ConfirmTransactionView ... perhaps they can be combined or you can just ask for confirmation to send (sweep) and then present a separate dialog box asking if the swept wallet should be deleted

lib/wallets/crypto_currency/coins/monero.dart
julian-CStack: Maybe a confirmation isn't required, but the mnemonic for the new wallet should be shown and confirmed by the user before the transaction is made. If anything goes wrong and the new wallet can't be accessed then the funds have been black holed

detherminal:
Added a dialog box for both mnemonic display and verify for the new wallet. Fixed other things as Tor, node choice and loading screen.

lib/utilities/address_utils.dart
```diff
@@ -174,6 +176,41 @@ class AddressUtils {
     }
   }
 
+  /// Parses a wallet URI and returns a WalletUriData object.
+  ///
+  /// Returns null on failure to parse.
+  static WalletUriData? parseWalletUri(String uri, {Logging? logging}) {
```
julian-CStack, line 182: Logging not used?

lib/utilities/address_utils.dart
```diff
@@ -10,6 +10,8 @@
 
 import 'dart:convert';
 
+import 'package:logger/logger.dart';
```
julian-CStack, line 13: Unintentional import? Doesn't appear to be used, if there are future plans to use logging in `parseWalletUri` then the Logging singleton in stack_wallet should be used instead of the logger package.

lib/wallets/crypto_currency/coins/monero.dart
```diff
@@ -124,6 +146,120 @@ class Monero extends CryptonoteCurrency {
     }
   }
 
+  @override
+  Future<Wallet<CryptoCurrency>?> importPaperWallet(WalletUriData walletData, WidgetRef ref, {Wallet? newWallet}) async {
```
julian-CStack, line 150: This is the wrong place for this function. I'm not sure where a good place would be but I do think this is the wrong place. I would say a static function in wallet.dart but static functions cannot be overridden in dart which is a weakness of `Wallet.create`. I think we need a wallet creation service that can be subclassed and thus have override functions. This would require a good chunk of refactoring that I do not have time for currently. Perhaps just doing stand alone functions would be a good short term solution. If done in a functional manner it should be relatively easy ti migrate to a wallet service class done in OOP style later on.

Thoughts?
cc @sneurlax

lib/pages/add_wallet_views/add_wallet_view/add_wallet_view.dart
```diff
@@ -125,6 +144,337 @@ class _AddWalletViewState extends ConsumerState<AddWalletView> {
...
+                        return results.coin.importPaperWallet(results, ref, newWallet: newWallet);
+                      })(),
+                      context: context,
+                      message: "Importing paper wallet...",
+                    );
+                    if (wallet == null) {
+                      await newWallet.exit();
+                      await ref.read(pWallets).deleteWallet(newWallet.info, ref.read(secureStoreProvider));
+                      if (mounted) {
+                        ScaffoldMessenger.of(context).showSnackBar(
```
julian-CStack, line 404: Either use the snackbar used elsewhere in stack wallet, or even better, show a dialog with some information about the error

lib/pages/add_wallet_views/add_wallet_view/add_wallet_view.dart
```diff
@@ -125,6 +144,337 @@ class _AddWalletViewState extends ConsumerState<AddWalletView> {
...
+                      await Navigator.of(context).pushNamed(
+                        WalletView.routeName,
+                        arguments: wallet.walletId,
+                      );
+                    }
+                  }
+                } else {
+                  await newWallet.exit();
+                  if (mounted) {
+                    ScaffoldMessenger.of(context).showSnackBar(
```
julian-CStack, line 425: Either use the snackbar used elsewhere in stack wallet, or even better, show a dialog with some information about the error

lib/pages/add_wallet_views/add_wallet_view/add_wallet_view.dart
```diff
@@ -125,6 +144,337 @@ class _AddWalletViewState extends ConsumerState<AddWalletView> {
...
+        } else {
+          if (mounted) {
+            final wallet = await showLoading(
+              whileFuture: results.coin.importPaperWallet(results, ref),
+              context: context,
+              message: "Importing paper wallet...",
+            );
+            if (wallet == null) {
+              if (mounted) {
+                ScaffoldMessenger.of(context).showSnackBar(
```
julian-CStack, line 448: Either use the snackbar used elsewhere in stack wallet, or even better, show a dialog with some information about the error

lib/pages/add_wallet_views/add_wallet_view/add_wallet_view.dart
```diff
@@ -125,6 +144,337 @@ class _AddWalletViewState extends ConsumerState<AddWalletView> {
...
+    if (kDebugMode) {
+      print("Mnemonic game correct word: $chosenWord");
+    }
+
+    return Tuple2(result, chosenWord);
+  }
+
+  Future<void> scanPaperWalletQr() async {
+    try {
+      final qrResult = await const BarcodeScannerWrapper().scan();
```
julian-CStack, line 181: camera/scanner logic has been updated everywhere else in the code base along with a permissions check. That should be used here as well

lib/wallets/crypto_currency/crypto_currency.dart
```diff
@@ -93,4 +104,40 @@ abstract class CryptoCurrency {
 
   @override
   int get hashCode => Object.hash(runtimeType, network);
+
+  Future<Wallet?> importPaperWallet(WalletUriData walletData, WidgetRef ref, {Wallet? newWallet}) async {
```
julian-CStack, line 108: Again, shouldn't be in this class

lib/wallets/crypto_currency/coins/monero.dart
```diff
@@ -124,6 +146,120 @@ class Monero extends CryptonoteCurrency {
...
+    try {
+      if (walletData.txids != null) {
+        // If the walletData contains txids, we need to create a temporary wallet to sweep the gift wallet
+        final wallet = await Wallet.create(
```
julian-CStack, line 154: What happens to this wallet and the files written to disk? What about if an error occurs at any point in the following code causing an early break in execution?

lib/pages/add_wallet_views/add_wallet_view/add_wallet_view.dart
```diff
@@ -125,6 +144,337 @@ class _AddWalletViewState extends ConsumerState<AddWalletView> {
...
+                              "Failed to import paper wallet for ${results.coin.prettyName}. Please try again.",
+                            ),
+                          ),
+                        );
+                      }
+                      return;
+                    }
+                    if (mounted) {
+                      Navigator.pop(context);
+                      await Navigator.of(context).pushNamed(
```
julian-CStack, line 416: Double check navigation works as expected when hitting the back button once in the wallet. Both desktop and mobile

lib/pages/add_wallet_views/add_wallet_view/add_wallet_view.dart
```diff
@@ -125,6 +144,337 @@ class _AddWalletViewState extends ConsumerState<AddWalletView> {
...
+                      "Failed to import paper wallet for ${results.coin.prettyName}. Please try again.",
+                    ),
+                  ),
+                );
+              }
+              return;
+            }
+            if (mounted) {
+              Navigator.pop(context);
+              await Navigator.of(context).pushNamed(
```
julian-CStack, line 460: Double check navigation works as expected when hitting the back button once in the wallet. Both desktop and mobile

lib/pages/add_wallet_views/add_wallet_view/add_wallet_view.dart
```diff
@@ -125,6 +144,337 @@ class _AddWalletViewState extends ConsumerState<AddWalletView> {
...
+            nodeService: ref.read(nodeServiceChangeNotifierProvider),
+            prefs: ref.read(prefsChangeNotifierProvider),
+            mnemonic: null,
+            mnemonicPassphrase: null,
+            privateKey: null,
+          );
+          await (newWallet as MoneroWallet).init(wordCount: 16);
+          final mnemonic = (await newWallet.getMnemonic()).split(" ");
+          if (mounted) {
+            final hasWroteDown = await showDialog<bool>(context: context, barrierDismissible: false, builder: (context) {
```
julian-CStack, line 205: Why not just use the mnemonic table code that is used when a regular new wallet is created?

julian-CStack (changes requested):
I'm probably missing some things as its been a long day

detherminal:
Moved the logic to a standalone function until wallet creation service is implemented.
Refactored SnackBar into app-wide FlushBar.
Added more error handling and cleaned.

detherminal:
@julian-CStack @sneurlax Sorry for ping.

PR #1160, opened 2025-07-16
Block explorer update for Banano and Nano

Verycutecat:
The block explorers for Banano and Nano are down. Replaced with Nanexplorer for Nano and Creeper for Banano.


julian-CStack:
I don't have experience using either explorer. Any reason to not just use nanexplorer for both banano and nano?

PR #1169, opened 2025-07-29
fix(macos): close mwebd on app close

sneurlax:


sneurlax:
Draft pending more testing.

sneurlax:
Tested working on Ubuntu

lib/main.dart
sneurlax, line 1: As suggested by @julian-CStack, `case AppLifecycleState.detached:` could probably just be `if (desktop) exit(0);`

julian-CStack:
I think monero (and friends) will also cause hanging occasionally

PR #1179, opened 2025-08-15
feat(test): add FFI integration testing suite to dev options menu

sneurlax:


sneurlax:
Haven't used this in a few months, but when I was, it was useful for quickly testing that all FFI libs work correctly.  I still feel that this is a useful function but need to find a better way to integrate it--until then, it'll just be a personal tool shown for those that're similarly interested 🦗

PR #1187, opened 2025-08-19
feat(i18n): localization/internationalization proof of concept

sneurlax:
this will take awhile to complete

let alone RTL layouts... this should maintain LTR layouts with RTL text therein, possibly-incorrectly LTR-justified, too.


sneurlax:
actually uses l10n (*ie.* localization)

PR #1188, opened 2025-08-21
feat(epic): Sync Epic Box transactions first

sneurlax:


sneurlax:
Converted to draft due to staleness.  Can probably close and delete branch safely

sneurlax:
This was created by request but never followed up on.  If we want something like this, we should re-examine the issue after the upcoming Epic updates.

PR #1227, opened 2025-12-08
Xelis 1.19 Compatibility

Tritonn204:
This PR migrates the Xelis integration in Stack to use the new FFI bindings for 1.19, which involved some straightforward (but still breaking) API changes and dep updates. This draft has been created to make the team aware of this pivot, while I re-sync to `staging` instead of `main`.

TODO:
- [X] Synchronize to `staging`
- [ ] ~~Walk back as many dep/build changes as possible~~ (`staging` sync eliminated the need)
- [X] ID any points of confusion in the devops AFTER syncing to `staging`
- [X] Sync my `xelis-dart-sdk` work to Slixe's parallel updates to fix current Genesix
- [X] Re-solidify the Xelis FFI dep tree to use the real endpoints instead of my forks
- [ ] ~~Make PR for downstream `cypherstack/xelis-flutter-ffi` w/ new API~~
- [X] Document the noteworthy mentionables for the team on this thread to allow asynchronous sync-ups


Tritonn204:
Dart `>= 3.1.0` is required, but docs pin the Flutter version to `3.35.7` which ships with only Dart `3.9.2`.

julian-CStack:
> Dart `>= 3.1.0` is required, but docs pin the Flutter version to `3.35.7` which ships with only Dart `3.9.2`.

Docs are out of date.  Will fix today

scripts/rust_version.sh
```diff
@@ -2,10 +2,10 @@
 
 
 set_rust_to_everything_else() {
-  if rustup toolchain list | grep -q "1.85.1"; then
-    rustup default 1.85.1
+  if rustup toolchain list | grep -q "1.91.0"; then
+    rustup default 1.91.0
```
julian-CStack: Does everything else in stack that uses this version of rust work as expected?

I actually forget everything that does so so I am hesitant to change this without looking at what uses this and testing those libraries. Everything should work but we've had issues in the past with some libraries that break when using newer versions of rust...

scripts/android/build_all.sh
```diff
@@ -10,8 +10,8 @@ PLUGINS_DIR=../../crypto_plugins
 # libepiccash requires old rust
 source ../rust_version.sh
 set_rust_version_for_libepiccash
-(cd "${PLUGINS_DIR}"/flutter_libepiccash/scripts/android && ./build_all.sh )
-(cd "${PLUGINS_DIR}"/flutter_libmwc/scripts/android && ./build_all.sh )
+# (cd "${PLUGINS_DIR}"/flutter_libepiccash/scripts/android && ./build_all.sh )
+# (cd "${PLUGINS_DIR}"/flutter_libmwc/scripts/android && ./build_all.sh )
```
julian-CStack: I believe these will need to be uncommented again. Probably known as this is still a draft PR but just in case...

scripts/rust_version.sh
Tritonn204: This will need to be combed through. Because xelis stuff is built on `flutter run`, I'm unsure whether we can restrict the scope to just that element. Admittedly, I haven't looked too deeply into the options regarding what as my focus has been elsewhere, but I haven't seen any issues in my brief testing.

Would it basically be Solana + Epic Cash that need smoke testing IF a better way of handling the rust needs for Xelis isn't found?

Tritonn204: *** I can see that epiccash sets to 1.85 and all else uses what was changed. Same question applies but just with the revised scope

Tritonn204: Phew, cargokit has a way apparently. I'll restore 1.85.1 but change the flutter FFI one to lock the cargokit toolchain, then revalidate.

PR #1240, opened 2026-01-09
Add Support for Firo Masternode Creation

cassandras-lies:
This PR adds support for Firo masternode creation. It doesn't support updating, so you need to break the masternode via coin control, send to self, and remake it for that. (Updating requires a BLS signature which uses a large outdated library I didn't want to pull in.)

This includes a DO NOT MERGE commit which includes https://github.com/julian-CStack/coinlib/pull/3 to support the masternode transactions. That should be merged upstream and the DO NOT MERGE commit rebased out.


julian-CStack:
cherry picked relevant commit and merged as part of https://github.com/cypherstack/stack_wallet/pull/1244

Issue #1211, opened 2025-10-25
Feature request: Linux: ability to force traditional titlebar


MorsMortium:
I have looked at [my_application.cc](https://github.com/cypherstack/stack_wallet/blob/a156253bf3494e094efffd332ee53ee65489157e/scripts/app_config/templates/linux/my_application.cc) and see that if the application is used on something other than Gnome and X11 is also used, it uses the traditional titlebar, in all other cases it uses the GTK CSD headerbar.
I can confirm that this does work, setting `WAYLAND_DISPLAY=""` does get back the titlebar.
But I would like to be able to have the titlebar on wayland, as a KDE user.
What I would recommend is also checking the environment variable GTK_CSD, and if it is equal to 0, then also disabling the headerbar.
Some other projects that implemented such changes:
https://github.com/stuartlangridge/magnus/pull/26/files
https://github.com/localsend/localsend/pull/669/files
https://github.com/midori-browser/core/pull/4/files
https://github.com/gpodder/gpodder/pull/791/files

julian-CStack:
seems easy enough

PR #1248, opened 2026-01-15
Don't use GNOME's custom title bar

pierrelasse:
Old ugly:
image

(Hopefully) fixed nicer looking:
image

Lmk if this is the correct file.

Issue about the issue: https://github.com/flutter/flutter/issues/111453

Here is someone else doing it https://github.com/interstellar-app/interstellar/commit/51956c3d702c574b4b06fb332158525d6185eae7

### DIFF
--- a/scripts/app_config/templates/linux/my_application.cc
+++ b/scripts/app_config/templates/linux/my_application.cc
@@ -20,32 +20,7 @@ static void my_application_activate(GApplication* application) {
   GtkWindow* window =
       GTK_WINDOW(gtk_application_window_new(GTK_APPLICATION(application)));
 
-  // Use a header bar when running in GNOME as this is the common style used
-  // by applications and is the setup most users will be using (e.g. Ubuntu
-  // desktop).
-  // If running on X and not using GNOME then just use a traditional title bar
-  // in case the window manager does more exotic layout, e.g. tiling.
-  // If running on Wayland assume the header bar will work (may need changing
-  // if future cases occur).
-  gboolean use_header_bar = TRUE;
-#ifdef GDK_WINDOWING_X11
-  GdkScreen* screen = gtk_window_get_screen(window);
-  if (GDK_IS_X11_SCREEN(screen)) {
-    const gchar* wm_name = gdk_x11_screen_get_window_manager_name(screen);
-    if (g_strcmp0(wm_name, "GNOME Shell") != 0) {
-      use_header_bar = FALSE;
-    }
-  }
-#endif
-  if (use_header_bar) {
-    GtkHeaderBar* header_bar = GTK_HEADER_BAR(gtk_header_bar_new());
-    gtk_widget_show(GTK_WIDGET(header_bar));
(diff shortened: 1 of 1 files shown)


julian-CStack:
The file is correct, however `main` branch is a ways behind `staging` and making this exact change there may break the change added to address https://github.com/cypherstack/stack_wallet/issues/1211 I think. 

I'm not familiar enough with how gtk works across various linux environments to say for sure whether this change will break things elsewhere. 

One more note: As this is a "template" file, the window title value should be left as 'PlaceHolderName' and https://github.com/cypherstack/stack_wallet/pull/1248/files#diff-293e8fb5d4eaabd128db2623823372c106b27c53a1e98a882ad639840713af83L101 should be left unchanged.

I will leave this PR open as it does look a lot better but it needs some tweaks and testing (or someone who knows more about gtk to chime in).

pierrelasse:
> I'm not familiar enough with how gtk works across various linux environments to say for sure whether this change will break things elsewhere.

~~From what I know, this *breaks* nothing. On GTK, it just makes the bar a bit smaller from what I know. (will edit this once I know it)~~
It should look like this on GNOME with this change.
image


> As this is a "template" file, the window title value should be left as 'PlaceHolderName' 

Where is this template used? So an application id can be set to improve support.

julian-CStack:
I did  a brief bit of digging and that `"application-id"` seems to be a key if I am not wrong and the value is `APPLICATION_ID` which is set in a cmakelists.txt. I followed that back to some config scripts and did find that we indeed did not properly set this. Thanks for bringing this up as I was able to fix it with https://github.com/cypherstack/stack_wallet/commit/0b2b797e3c6e0f092943dbc34bda12de14e6b5fd

Regarding the title bar, the title string is configured before compile time so that should not be changed here. As I mentioned previously, I agree with the better looks of the smaller/thinner title bar and the change #WorksOnMyMachine. But not everyone runs Ubuntu with GNOME and we've had various issues in the past with different distros, so without testing on a bunch of different systems this is a change I'm hesitant to add at the moment.

pierrelasse:
Can you maybe approve the workflow so we can get a version we can perform these tests with?

julian-CStack:
I've updated the base branch according to my first message re `staging` branch. I think the workflow will fail currently regardless (due to unrelated issues) but that won't stop you from building a version to test with.

pierrelasse:
womp womp

sneurlax:
@pierrelasse no worries, my personal github has raw, unfinished work: when I get around to cleaning it up, I'll redo it and base it on your work here so your contribution is recognized officially.

pierrelasse:
Great. You can also add me as Co Author if that makes it easier for you to manage

sneurlax:
@pierrelasse your work was cherry-picked directly onto fix/gnome and included in https://github.com/cypherstack/stack_wallet/pull/1328 .  This PR will be closed sometime after that's merged.  Thanks for your contribution!

julian-CStack:
merged https://github.com/cypherstack/stack_wallet/pull/1328

PR #1270, opened 2026-03-03
fix: show token ticker in DesktopAuthSend prompt

sneurlax:
Came across this issue while working on ShopInBit's USDT payment method: I noticed that the DesktopAuthSend dialog showed ETH and not USDT and fixed this there, but this adds an optional tokenTicker param to all regular dialogs and uses it from ConfirmTransactionView


julian-CStack:
I think we should add the parent chain info as well if its a token. We do that elsewhere in the app and I think its called coin tag or something...

PR #1271, opened 2026-03-03
Fix MWC: fix global chain type race and use auth for mwc713.mwc.mw

sneurlax:
2 fixes in one, one which fixes syncing and another which fixes a crash bug


lib/wallets/wallet/impl/mimblewimblecoin_wallet.dart
```diff
@@ -585,6 +593,21 @@ class MimblewimblecoinWallet extends Bip39Wallet {
     return stringConfig;
   }
 
+  /// Write the node API secret to .api_secret in the wallet directory so that
+  /// the Rust HTTPNodeClient can authenticate to the MWC node.
+  Future<void> _ensureApiSecret(String walletDir, String nodeUrl) async {
+    const defaultNodeHost = 'mwc713.mwc.mw';
+    const defaultNodeSecret = '11ne3EAUtOXVKwhxm84U';
```
sneurlax: We should probably unify this secret so the MWC node test file doesn't have to reproduce it (or vice versa).  A cleaner approach may be to add an optional auth field/param to the NodeModelAdapter, but I didn't want to get into all that right now.

danrmiller (approved):
LGTM. Tested on linux.

lib/wallets/wallet/impl/mimblewimblecoin_wallet.dart
```diff
@@ -894,14 +917,17 @@ class MimblewimblecoinWallet extends Bip39Wallet {
...
         //Open wallet
-        encodedWallet = await libMwc.openWallet(
-          config: stringConfig,
-          password: password,
-        );
-        await secureStorageInterface.write(
-          key: '${walletId}_wallet',
-          value: encodedWallet,
-        );
+        encodedWallet = await _walletOpenMutex.protect(() async {
```
danrmiller: What if we just call _ensureWalletOpen and use the mutex you already added there?

lib/wallets/wallet/impl/mimblewimblecoin_wallet.dart
```diff
@@ -935,14 +961,16 @@ class MimblewimblecoinWallet extends Bip39Wallet {
...
 
-          final walletOpen = await libMwc.openWallet(
-            config: config,
-            password: password!,
-          );
-          await secureStorageInterface.write(
-            key: '${walletId}_wallet',
-            value: walletOpen,
-          );
+          await _walletOpenMutex.protect(() async {
```
danrmiller: Same question, should we just call _ensureWalletOpen or is this better, adding another mutex?

lib/wallets/wallet/impl/mimblewimblecoin_wallet.dart
```diff
@@ -585,6 +593,21 @@ class MimblewimblecoinWallet extends Bip39Wallet {
...
   }
 
+  /// Write the node API secret to .api_secret in the wallet directory so that
+  /// the Rust HTTPNodeClient can authenticate to the MWC node.
+  Future<void> _ensureApiSecret(String walletDir, String nodeUrl) async {
+    const defaultNodeHost = 'mwc713.mwc.mw';
+    const defaultNodeSecret = '11ne3EAUtOXVKwhxm84U';
+
+    final file = File('$walletDir/.api_secret');
+    if (nodeUrl.contains(defaultNodeHost)) {
```
danrmiller: I recently had something come up totally unrelated, where example.co would match where only example.com was intended. Is this a similar issue, for example evil.mwc713.mwc.mw.attacker.com?

lib/wallets/wallet/impl/mimblewimblecoin_wallet.dart
danrmiller: Another question, how is the mutex disposed of or timed out?

danrmiller (dismissed):
I'm approving this as it improves the current state of the app, mwc will now work with this merged.

sneurlax:
Closes #1303 now too

lib/wallets/wallet/impl/mimblewimblecoin_wallet.dart
sneurlax: Resolved by 22a609f8853fe8b97880ec0ed09e1c47fbe47b98

lib/wallets/wallet/impl/mimblewimblecoin_wallet.dart
sneurlax: Resolved by 22a609f8853fe8b97880ec0ed09e1c47fbe47b98

lib/wallets/wallet/impl/mimblewimblecoin_wallet.dart
sneurlax: Resolved by 6170e8cd57aaac054975ea6adbe6f238441e1166 and the trail of commit leading up to that on #feat/node-api-secret-in-model

sneurlax:
Have to recheck that 

- [x] users can change the password in settings if mwc changes the password on their node

lib/wallets/wallet/impl/mimblewimblecoin_wallet.dart
```diff
@@ -95,32 +97,43 @@ class MimblewimblecoinWallet extends Bip39Wallet {
...
-      throw Exception('Wallet password not found');
-    }
-    final opened = await libMwc.openWallet(config: config, password: password);
-    await secureStorageInterface.write(
-      key: '${walletId}_wallet',
-      value: opened,
-    );
-    return opened;
+      // Drop stale pointer left by pre-instance-var builds.
+      await secureStorageInterface.delete(key: '${walletId}_wallet');
```
julian-CStack: This only needs to be run once. Since there is already a migration added in this PR, it would be better to put this there I think.

lib/wallets/wallet/impl/mimblewimblecoin_wallet.dart
```diff
@@ -576,6 +592,8 @@ class MimblewimblecoinWallet extends Bip39Wallet {
     final String nodeApiAddress = uri.toString();
     final walletDir = await _currentWalletDirPath();
 
+    await _ensureApiSecret(walletDir);
```
julian-CStack: Should this be called in updateNode()  instead of here?

lib/wallets/wallet/impl/mimblewimblecoin_wallet.dart
sneurlax: Agreed, fixing.

sneurlax: Resolved by af57657fb4e9be79cd273c73fed0b4c7f4329779

lib/wallets/wallet/impl/mimblewimblecoin_wallet.dart
sneurlax: Resolved by e99bbc0e16c4dfd72576f7235d110383901c3691

julian-CStack (approved):

sneurlax:
followup comment: "merged. I did notice that the wallet handle var was directly accessed in places vs being ensureOpened but I think that is due to checks if the wallet is not open which should be fine as it replaces what I think was current functionality anyways. May be something to look into in the future to see if it could be hanlded a bit differently/safer"

Issue #6, opened 2022-08-28
Provide reproducible builds


AaronFeickert:
Users should not have to trust that Stack Wallet binaries properly represent the source tree. Reproducible builds mitigate this risk by allowing users to check that a given binary matches a given commit.

mvozzo:
I support @AaronFeickert comment above.  
This will help mitigate unwarranted code injections by hackers / bad actors within Big Tech. Big brother is watching and with Big Tech working with Big Banks and the eminent rollout of CDBCs, wallets like Stack Wallet may be targeted.

PR #1274, opened 2026-03-04
Linux x86_64 reproducible builds

sneurlax:
Android might be feasible and I am checking the other platforms but we should only expect success for Linux and Android

~~I am updating build_app.sh to add a guix option and I am also updating the stack_releases scripts to make it easier to use this~~


sneurlax:
Closes https://github.com/cypherstack/stack_wallet/issues/6

IF the supported platforms are enough AND we can't figure out the rest of the platforms.  Maybe guix will work in WSL2 for Windows builds and maybe if not guix then nix or something like that can work for macOS and/or iOS (I have doubts about the feasibility for iOS...)

sneurlax:
cs_monero linux guix proof of concept pushed to https://github.com/cypherstack/cs_monero/compare/feat/guix ... but guix won't be suitable for macos and ios and there, nix will have to be used

nix is compatible w linux

first I am just seeing if nix can prove reproducibility for cs_monero

if that works then stack wallet may see a feat/nix instead of a feat/guix and this will just be an example

sneurlax:
got nix working on linux and the rest of the platforms san win but just for flutter, not for stack wallet. TODO

sneurlax:
see progress on https://github.com/cypherstack/stack_wallet/tree/feat/stagex-reproducible-builds ... need to 'just' incorporate monero's new stagex tooling for monero_c->cs_monero, too

PR #1275, opened 2026-03-05
Fix throw when toggling Tor with unopened cryptonote family wallets

sneurlax:
When you toggle Tor with an uninitialized cryptonote wallet, it throws.  This isn't fatal.  The fix uses the same guard pattern other functions use.

This fix is admittedly optional: it's not a critical fix but rather a cosmetic/log-cleaning fix.

### DIFF
--- a/lib/wallets/wallet/intermediate/cryptonote_wallet_lifecycle.dart
+++ b/lib/wallets/wallet/intermediate/cryptonote_wallet_lifecycle.dart
@@ -1,0 +1,138 @@
+import 'dart:async';
+
+import 'package:mutex/mutex.dart';
+
+Future<void> cancelCryptonoteWalletSubscriptions(
+  Iterable<StreamSubscription<dynamic>?> subscriptions,
+) async {
+  Object? firstError;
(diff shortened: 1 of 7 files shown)


sneurlax:
Comments:
"""
does this null check https://github.com/cypherstack/stack_wallet/pull/1275/changes hide a problem elsewhere? 
I'm looking at 1275 again. If the null check hides a problem elsewhere and that problem was fixed, why is the null check still there?
any chance introducing another mutex (_updateNodeMutex) could cause a deadlock?
Ok so I am generally in favor of writing less code and doing the same check in a single place instead of spreading it all over the place but I still think the null check is going a bit too far in catching every single call.

For example in https://www.github.com/cypherstack/stack_wallet/blob/211971b5d7c6cac19173071026388dbaf756e8ab/lib/pages/settings_views/global_settings_view/manage_nodes_views/add_edit_node_view.dart#L306-L339
The check should probably be handled here as refresh is called on the same wallets. Maybe I'm overthinking it though?
"""

PR #1277, opened 2026-03-05
fix: _getLastAutoBackup() reading from wrong Hive key

sneurlax:


julian-CStack (approved):

julian-CStack:
Looks like this is what it should be but I have not tested

PR #1261, opened 2026-02-22
feat: support monero_wallet URI restoration

sneurlax:
supports seed, view_key only, and spend_key and view_key.

PR #1278, opened 2026-03-05
Feat/restore cryptonote from date or height

sneurlax:
Gated to cryptonote and mimblewimble
Screenshot from 2026-03-04 22-35-07
 coins


sneurlax:
Converted to draft because this needs to be merged after #1261

PR #1286, opened 2026-03-19
Fix default Solana token list and support `transfer` (not just `transferChecked`)

sneurlax:


danrmiller (dismissed):
I think the token transaction parsing is missing the "innerInstructions" and those get missed.

I'm also not sure about the splTransfers.length != 1, I'm pretty sure that is going to miss things on multi-recipient transactions.

But those can be addresses seperately.

PR #1289, opened 2026-03-23
Migrate from PayNym.is to PayNym.rs

sneurlax:
and gracefully handle unexpected responses
and add tests


sneurlax:
tested linux, need to test all other platforms

PR #1295, opened 2026-04-07
Optimize Firo Spark mint tx generation and fix fee < vSize error

reubenyap:
## Summary

Two changes to `_createSparkMintTransactions` in `spark_interface.dart`:

1. **Performance**: Pre-compute signing keys, addresses, and wallet-owned address set once before the main loop, instead of per-UTXO per-fee-iteration.
2. **Bug fix**: Scale the fee-estimation buffer with input count to avoid the `fee is less than vSize` error on many-input mints (e.g., anonymize-all).

## Performance optimization

Hot-loop operations moved out of the nested loops:

| Operation | Before | After |
|---|---|---|
| `getRootHDNode()` | per UTXO × per fee iteration | once |
| `mainDB.getAddress(walletId, ...)` | per UTXO × per fee iteration | once per unique address |
| `mainDB...walletIdEqualTo(...).valueEqualTo(...).findFirst()` for `walletOwns` | per output × per fee iteration | once, as a `Set<String>` lookup |
| `getCurrentReceivingSparkAddress()` / `getCurrentChangeAddress()` | per inner iteration | once |

For an N-input, M-fee-iteration mint this eliminates ~N×M expensive async calls.

## Fee bug fix

**Root cause.** The dummy tx built for fee estimation is signed with the real keys over *different data* than the final tx. `bitcoindart`'s ECDSA path (`bip32-dart/ecurve.sign`) uses RFC 6979 and enforces low-S but **not low-R**, so `r`'s high bit is random — DER signatures vary by up to ~4 bytes per input. For Firo P2PKH (bip44) inputs this counts at full weight; with ~10+ inputs the dummy-vs-real vSize delta exceeds the original 10-byte buffer and trips `nFeeRet < data.vSize`.

**Fix.** Scale the buffer with input count:

```dart
final nBytesBuffer = 10 + 4 * setCoins.length;
```

Extra cost: ~4 sats per input at 1 sat/byte. Matches what Firo's own C++ wallet does — see `DummySignatureCreator` in [`src/wallet/wallet.h:1436`](https://github.com/firoorg/firo/blob/master/src/wallet/wallet.h#L1436) ("max-sized low-S signatures (eg 72 bytes)" → fee estimation is always an upper bound).

## Subsidiary fixes

- `mintedValue <= BigInt.zero` (was `== BigInt.zero`): catches negative `mintedValue` when a UTXO group's total is less than the computed fee and `subtractFeeFromAmount=false`. Matches the C++ reference's `!MoneyRange(mintedValue) || mintedValue == 0`.
- Clarified the `fee < vSize` error message and added a comment: the check is a disguised min-relay-fee check (`feeRate < 1 sat/byte`), not a fee/size accounting error.

## Why not retry?

An earlier iteration used a retry loop that restored UTXOs, outputs, and fee state when the check fired. Investigation confirmed the size variance has a deterministic upper bound (ECDSA DER max per input), so a scaled buffer is cleaner — no state to restore, no wasted mint-proof generation, no risk of loop bugs.

## Test plan

- [ ] Single-recipient Spark mint (1–3 inputs): verify no regression.
- [ ] Anonymize-all with 10+ small UTXOs: verify the tx builds without `fee is less than vSize`.
- [ ] Sanity check: signing-key cache miss (corrupted DB) throws a clear rescan message.


reubenyap:
Post-review note after a second pass:

Additional hardening pushed:
- `10ced8df2`: fee subtraction now keeps redistributing after tiny Spark outputs are removed, then verifies actual input-output fee equals `nFeeRet`.
- `b8f66850d`: auto-mint-all now throws if no Spark mint transactions are produced, so future empty-result paths cannot confirm as success.

Second-pass checks:
- PR diff is still one file: `lib/wallets/wallet/wallet_mixin_interfaces/spark_interface.dart`.
- Rechecked the signing-key cache assumption: Firo supports only `DerivePathType.bip44`, and `FiroWallet` does not mix in `PaynymInterface`.
- Re-ran the pathological fee case `[100, 100, 1]` with fee `150`: outputs become `25, 26`, paid fee `150`, remaining fee `0`.
- Re-ran an exhaustive small sweep for three-output fee subtraction (`1..12` output values, `1..36` fees): `0` failures.
- `git -c core.autocrlf=false diff --check` passes.

I did not find remaining bugs/regressions in the current one-file diff. Local limitation: `dart`/`flutter` are not on PATH in this environment, so I could not run the project test suite locally.

julian-CStack (approved):

PR #1304, opened 2026-04-27
chore: dart format staging vs main

sneurlax:
`dart format $(git diff --name-only origin/main...HEAD | grep '\.dart$')`


julian-CStack:
Formatted code is satisfying but I think it would be better to keep as is until the files are modified and then handle auto format. Can revisit later

PR #1308, opened 2026-04-29
Add -s flag to build_app.sh to use system jsoncpp and libsecret

danrmiller:


sneurlax (dismissed):
LGTM and I tested this and it apparently works both ways.  But I didn't test in a fresh env which might fail with -s

danrmiller:
> LGTM and I tested this and it apparently works both ways. But I didn't test in a fresh env which might fail with -s

We probably want it to fail if the libs aren't installed and -s is used. We could try to detect them but I think this keeps everything in line with existing expectations.

PR #1328, opened 2026-05-07
feat: drop GNOME's custom title bar, allow to be used as in #1211

sneurlax:
Includes and supercedes https://github.com/cypherstack/stack_wallet/pull/1248

Inverts the env var introduced in https://github.com/cypherstack/stack_wallet/pull/1248 from opt-out to opt-in


julian-CStack (approved):
lets see what happens

PR #1342, opened 2026-05-16
Add OP_RETURN Support Required for Rosen Bridge

navidR:
This PR adds `OP_RETURN` output support to Campfire wallet’s Firo (Firo) transparent send flow, enabling users to participate in Rosen Bridge transfers by pasting bridge payment URIs. When a URI containing the op_return parameter is pasted or scanned, the wallet automatically extracts the hex payload, attaches it as an `OP_RETURN` output with proper push‑data encoding (supporting payloads up to 80 bytes using `OP_PUSHDATA1` for lengths 76–80), and includes the extra output in all fee‑estimation and coin‑selection steps. The feature is currently scoped to Firo’s transparent‑to‑transparent transactions only; Spark and MWEB privacy transactions are unaffected and will not carry `OP_RETURN` data.

A safety guard has been added to prevent loss of funds when a user pastes a bridge URI while the wallet is set to a private (Spark) balance. In that scenario, a warning message is displayed and the send/preview button is disabled, ensuring that bridge data is never silently omitted from a transaction. The guard clears automatically when the user switches to the public balance or clears the send form. The patch has been reviewed for script correctness, fee estimation accuracy, and internal state cleanliness (`OP_RETURN` data resets on form clear), making the wallet fully ready for Rosen Bridge transparent Firo transfers with no risk of mis‑bridging.

### DIFF
--- a/lib/pages/send_view/send_view.dart
+++ b/lib/pages/send_view/send_view.dart
@@ -257,6 +257,7 @@ class _SendViewState extends ConsumerState<SendView> {
             paymentData.coin?.uriScheme == coin.uriScheme) {
           _applyUri(paymentData);
         } else {
+          _setOpReturnData(null);
           if (coin is Epiccash) {
             content = AddressUtils().formatEpicCashAddress(content);
           }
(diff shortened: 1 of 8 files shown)


julian-CStack (approved):
I think this is functionally ok

PR #1344, opened 2026-05-18
fix windows build doc

Cyrix126:


docs/building.md
```diff
@@ -149,13 +149,21 @@ cd scripts/windows
...
+
+install go in WSL [https://go.dev/doc/install](https://go.dev/doc/install) (follow linux instructions) and ensure you have `mingw-w64` package installed to get the `x86_64-w64-mingw32-gcc` compiler.
+
 go version should be at least 1.24
 
 and use `scripts/build_app.sh` to build plugins:
 ```
 cd ..
-./build_app.sh -a stack_wallet -p windows -v 2.1.0 -b 210
+./build_app.sh -a stack_wallet -p windows -v 2.4.4 -b 301
```
julian-CStack, line 166: Only thing I'd change is putting some kind of place holder or variable for the version and build numbers. These change and can also cause issues on some platforms if someone runs a version that is older than the currently installed version. I don't think this applies to windows though

julian-CStack (commented):
LGTM

docs/building.md
Cyrix126: This section https://github.com/cypherstack/stack_wallet/blob/staging/docs/building.md#build-script-build_appsh does exactly that but it's in the linux chapter. I'll make a reference to it. Should be enough ?

julian-CStack (approved):

PR #1356, opened 2026-05-27
auto-label UTXOs from transaction notes

### LINKED ISSUE #411: Feature Request: Auto Labelling UTXOs off Transaction Notes
I noticed when I add a note to a transaction that the UTXOs remaining in the wallet from that tx don't inherit the note as a label. This is a nice feature that exists in Samourai and Sparrow, and saves a little time when labelling UTXOs.

I would assume a Blank UTXO label should inherit a TX note. If the UTXO already has a label then that label should persist even if the tx note changed.

sneurlax:
closes #411


sneurlax:

PR #1365, opened 2026-05-27
feat: show sat/vB fee rate inline next to fee amount

### LINKED ISSUE #413: UI Request: Sat/vB visiable next to Fee Amount
On the fee rate portion of the Bitcoin Wallet transaction creation screen I would like the sats/vB number to be visible to the total fee amount.
This fee rate number should always be in sats/vB regardless of whether you show the wallet denominations in BTC or Sats.

sneurlax:
closes #413


sneurlax:

PR #1371, opened 2026-05-27
feat: portable AppImage mode detection

### LINKED ISSUE #1177: Add portable aware storage code for portable mode with AppImages
So AppImages supports portable mode and some AppImages of other projects have `--portable` flag.
I tried `./ sw-v2.3.2-linux.AppImage --portable` but it still created a directory in home `.stackwallet`, so it didn't work. 

Turns out that command is not an automatic switch for AppImages but rather logic added to the code of an AppImage?

[Using Portable Mode Documentation](https://docs.appimage.org/user-guide/portable-mode.html)

Reading pull request #1091 it might be a good idea to implement OS detection for this. You could indeed implement logic in your Flutter/Dart application to check the operating system name and initiate the `--portable` option if it matches a certain condition. This can be done using platform checks available in Dart. Although I'm not using Whonix or Tails rather just Debian. 

My use case is I would like to put my Stack wallet AppImage binary in my encrypted partition in a way that all Stack Wallet Data and settings are contained in the folder the binary sits in inside the LUKS partition. Going off what said previously it would be good for those two OS's to work seemlessly for this. This could also be expanded to a setting in Advanced settings for "Enable Portable Mode". So if the logic detects those two OS's as mentioned in #1091 it would turn that setting on. Otherwise the user would have to turn it on manually if they are not on those systems.

sneurlax:
closes #1177

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -24,6 +24,7 @@ import 'package:google_fonts/google_fonts.dart';
 import 'package:keyboard_dismisser/keyboard_dismisser.dart';
 import 'package:logger/logger.dart';
 import 'package:mobile_app_privacy/mobile_app_privacy.dart';
+import 'package:path/path.dart' as path;
 import 'package:path_provider/path_provider.dart';
 import 'package:window_size/window_size.dart';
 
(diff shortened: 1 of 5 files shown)


sneurlax:

PR #1375, opened 2026-05-27
fix: WillPopScope migration

sneurlax:
Migrate the deprecated WillPopScope widget to PopScope across the app. Simple barrier dialogs become canPop: false/true. Cases that ran side effects or conditionally gated the pop move that work into onPopInvokedWithResult.


sneurlax:

PR #1378, opened 2026-05-27
feat: coin control row tap-to-toggle and per-address options menu

### LINKED ISSUE #406: Coin control menu address improvements
I would like to suggest the following features for the addresses in the coin control menu

- Clicking on the address does go into the address options, instead it will toggle checking/unchecking the address for coin control
- An "..." button is on the right side of the address (Just like how the checkbox is on the left side of the address) which will take you to the address options.

sneurlax:
Tapping an address/output row now toggles its selection instead of opening the details view. A vertical ellipsis options button is added to the right of each row to reach the address options. Keeps the filtering, WillPopScope migration, and the UTXOConfirmedStatus deduplication from the prior work.

closes #406


sneurlax:

PR #1380, opened 2026-05-27
fix(logger): avoid infinite recursion when isolate log port fails

sneurlax:


julian-CStack (changes requested):
I agree this could be improved. But I do not thick these requested changes are the solution.

I've talked to someone before about this exact spot but I cannot remember who.

sneurlax:
We talked about this but you might've discussed it with someone else.  I made 374c7b8, then you revisited the same block in 2887dc1 and 8c3f6e7.  So I see that this is basically a reversion to something you already nACKed and will take another look at it...

sneurlax:
OK, errors are typically written to [`latest.txt`](https://github.com/cypherstack/stack_wallet/pull/1380/changes#diff-d9fce43e1fca51ee54e33aa87eaca793b9e82b3d5177b2537351c6c14bb52dc6R101), with [`emergency.txt`](https://github.com/cypherstack/stack_wallet/pull/1380/changes#diff-7bee801003c0212b28ee00fbc3f23260132cb7dbfb4303ce953b9714a4c922cdR33) used if the former doesn't work, in a portable (cross-platform) way.  I still need to test this on the rest of the platforms:

- [x] Linux
- [x] macOS
- [ ] Windows
- [x] Android
- [x] iOS

PR #1381, opened 2026-05-27
Replace bitcoindart and coinlib_flutter with the coin library

sneurlax:
Migrate all UTXO key derivation, address generation, network parameters, transaction building, signing, and message signing in the wallet engine from bitcoindart/coinlib_flutter to the coin library, added as a submodule under crypto_plugins/coin and wired as a path dependency.

Coin definitions and the Bip39HD currency/wallet bases now use coin.Chain and coin key/address/script types. The ElectrumX, PayNym, Spark, MWEB, CashFusion, and extended-keys interfaces are restructured onto coin sign-then-assemble transaction building. PayNym gains P2TR payment code support. Particl retains its witness-only signing format, and Firo ProReg collateral signing moves to coin MessageSig.

Fork bip47 and fusiondart locally under crypto_plugins so their internals use coin, and drop the direct bitcoindart and coinlib_flutter dependencies and overrides. Add bitcoindart-coverage and coin equivalence/regression test suites.


sneurlax:
This doesn't need to have coin added as a submodule, it's already published.  I should also publish the bip47 and fusiondart changes on branches on either my sneurlax or the main cypherstack remote rather than moving them out into crypto_plugins submodules.  TODO

Issue #1399, opened 2026-06-03
Linux AppImage v2.4.4: frostdart.so absent from official release


xrviv:
## Context

[WalletScrutiny](https://walletscrutiny.com) is an independent security research project that verifies whether Bitcoin and cryptocurrency wallet binaries can be reproduced from their public source code. We recently completed reproducible build verification for the Stack Wallet **Linux AppImage v2.4.4** (`sw-v2.4.4-linux.AppImage`, SHA256 `ae6f3f31c2c894384aa54bd481506ee6cc133a8de1d61aa18e96722b8d1c8809`).

During this work we identified a discrepancy between what the build system produces and what is packaged in the official AppImage.

---

## Finding

The official `sw-v2.4.4-linux.AppImage` **does not contain `frostdart.so`** in its `lib/` directory.

We confirmed this by extracting the squashfs from the AppImage and listing its contents:

```
unsquashfs -o 944632 -d squashfs-root sw-v2.4.4-linux.AppImage
ls squashfs-root/lib/ | grep frost
# (no output)
```

When we build from the public source at commit `edc19aa6a23052de854be0318190f82bdcc0a765` (the commit embedded in the official binary's `libapp.so`), with Rust 1.71.0 installed, the build produces `frostdart.so` at:

```
crypto_plugins/frostdart/scripts/linux/build/frostdart.so
```

The frostdart Flutter plugin's `linux/CMakeLists.txt` unconditionally installs this file into the bundle:

```cmake
install(FILES "${CMAKE_CURRENT_SOURCE_DIR}/../scripts/linux/build/frostdart.so"
        DESTINATION "${INSTALL_BUNDLE_LIB_DIR}"
        COMPONENT Runtime)
```

There is no `OPTIONAL` flag and no `if(EXISTS ...)` guard on this install directive. Our local build includes `frostdart.so` in the bundle; the official AppImage does not.

---

## Related: Rust 1.71.0 not documented

`crypto_plugins/frostdart/scripts/linux/build_all.sh` hardcodes:

```bash
cargo +1.71.0 build --release ...
```

However, `docs/building.md` lists only `rustup install 1.89.0 1.85.1 1.81.0` — Rust 1.71.0 is not mentioned anywhere in the documentation. Anyone following the official build instructions exactly will fail to compile `frostdart` for Linux.

---

## Questions

[...]

PR #1400, opened 2026-06-03
fix(docs): correct Flutter version, Rust toolchains, Ubuntu 24.04 meson install, add rsync

xrviv:
## Summary

Corrections to `docs/building.md` found during [WalletScrutiny](https://walletscrutiny.com) reproducible build verification of Stack Wallet v2.4.4 (see issue #1399).

- **Flutter version**: `3.38.5` → `3.38.1` on all platforms (Linux, macOS, Windows). Confirmed by `libflutter.so` identity match against the official v2.4.4 binary.
- **Rust 1.71.0**: Added to all `rustup install` lines. `crypto_plugins/frostdart/scripts/linux/build_all.sh` hardcodes `cargo +1.71.0`; without it, frostdart fails to compile on all platforms.
- **Ubuntu 24.04 meson install**: Replaced `pipx` with a `python3-venv` approach. `pipx` cannot install library-only packages (`markupsafe`, `jinja2`, etc.) and will exit with an error on Ubuntu 24.04. `setuptools` is added because Python 3.12 removed `distutils`, which meson requires for the libsecret GObject introspection step.
- **`rsync`**: Added to Linux desktop deps. `crypto_plugins/flutter_libmwc/scripts/linux/build_all.sh` uses `rsync` to copy source into the build directory before running `cargo`; without it the build fails immediately.

## Test plan

- [ ] Confirm `flutter --version` reports `3.38.1` after following updated Flutter install step
- [ ] Confirm `cargo +1.71.0 --version` works after following updated rustup install step
- [ ] Confirm `meson --version` reports `0.64.1` after following updated Ubuntu 24.04 meson install step
- [ ] Confirm `scripts/linux/build_app.sh -p linux` completes without rsync error

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -14,7 +14,7 @@ Here you will find instructions on how to install the necessary tools for buildi
 The following instructions are for building and running on a Linux host.  Alternatively, see the [Mac](#mac-host) and/or [Windows](#windows-host) section.  This entire section (except for the Android Studio section) needs to be completed in WSL if building on a Windows host.
 
 ### Flutter
-Install Flutter 3.38.5 by [following their guide](https://docs.flutter.dev/get-started/install/linux/desktop?tab=download#install-the-flutter-sdk).  Run `flutter doctor` in a terminal to confirm its installation.
+Install Flutter 3.38.1 by [following their guide](https://docs.flutter.dev/get-started/install/linux/desktop?tab=download#install-the-flutter-sdk).  Run `flutter doctor` in a terminal to confirm its installation.
 
 ### Android Studio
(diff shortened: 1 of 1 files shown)


danrmiller (commented):
The macOS 1.71.0 addition is right, frostdart's own CI does install 1.71.0 on macOS. But installing it alone isn't enough: `crypto_plugins/frostdart/scripts/macos/build_macos.dart` calls `cargo lipo` with no toolchain override, so the local build still uses whatever `rustup default` is (1.89.0 per these docs). The upstream CI works around this by setting 1.71.0 as the active toolchain before invoking cargo. The local script would need `cargo +1.71.0 lipo ...` to actually match. That's a pre-existing issue in the submodule script, not introduced here, worth a follow-up on the frostdart side.

Everything else looks good. Thanks for the thorough investigation.

PR #1407, opened 2026-06-14
Change fact0rn explorer address from non-working one.

danrmiller:
Per fact0rn devs, https://factexplorer.io/ is gone for good, and https://explorer.fact0rn.io/ displays unconfirmed mempool transactions.


julian-CStack (approved):
who needs mempool transactions if there is no other explorer

PR #1415, opened 2026-07-15
feat: add OpenCryptoPay support

Cyrix126:
Some notes:

feat: refactor some part of lib/pages/send_view/confirm_transaction_view.dart to avoid dupplicating code
feat: add an http client class that can be used for library needing a Client from the http package as parameter
fix: Add missing qr scan button on the token view for desktop


reubenyap (commented):
Reviewed `2bba7deb` against `staging` at `9517439b`. I revalidated each finding; this still needs changes before merge:

- `opencryptopay` is pinned to missing commit `adfc0d1fb4620e84fcd3833a7cb99570770eeb76`. CI and a clean `flutter pub get` fail, and current package `main` is not API-compatible.
- The OCP session is re-associated solely by recipient address while the amount remains editable. For txid-proof methods, an edited form can be broadcast before proof is rejected. Bind or invalidate the session using the quoted recipient and exact amount.
- The txid-proof path broadcasts first, starts `submitProof` unawaited, then navigates away. A delayed failure can become silent, and on mobile the retained retry state is disposed. Await proof and preserve a durable txid-only retry path that cannot rebroadcast.
- Signed-hex submission treats a timeout or lost response as “nothing was broadcast,” even if the provider already broadcast. In UTXO flows, inputs remain reusable. Quarantine before submission, then reconcile the result.
- OCP amounts are round-tripped through display text. In SPL views, a parseable non-normal SOL unit can turn 0.5 tokens into 500; in ERC-20 views, a low `maxDecimals` can turn 1.246858 into 1.24. Keep the parsed `Amount` as the transaction input.
- The protocol’s per-method `minFee` is discarded, while tests cover only `minFee: 0`.
- An accepted ETH/ERC-20 hex payment creates no immediate local pending record and does not save its note. For UTXO hex payments, a post-accept DB error can leave the non-dismissible progress dialog open and skip completion.
- For a QR-selected endpoint, `StackClient.timeout` does not cancel the source request, and the response stream is buffered without a size limit.

`git diff --check`, formatting, and current-base merge integration pass. Exact tests remain blocked by the missing dependency. I found no issue with the GET-only protocol calls or EIP-1559 signing prefix.

julian-CStack:
I merged this into https://github.com/cypherstack/stack_wallet/tree/julian/ocp since I believe at least one of the points Reuben posted is addressed by changes I made in another branch that forms the base of the julian/ocp branch.

I will leave this PR open for now

Cyrix126:
> I merged this into https://github.com/cypherstack/stack_wallet/tree/julian/ocp since I believe at least one of the points Reuben posted is addressed by changes I made in another branch that forms the base of the julian/ocp branch.
> 
> I will leave this PR open for now

Indeed the ocp branch has reworked min fees so it's better to work from it. I'm merging ocp into this branch and fixing the rest that is relevant from here.

PR #1422, opened 2026-08-04
Fix Firo OP_RETURN fee selection

navidR:
Campfire calculated Firo transaction fees without including the Rosen OP_RETURN output, so the final transaction could be underfunded and fail. This fix includes the OP_RETURN output in the fee calculation. A regression test covers 75-byte Rosen metadata and confirms normal Firo transactions are unchanged.


julian-CStack:
This should be handled with the fee calc changes in staging now

PR #1423, opened 2026-08-05
chore(xelis): bump xelis_flutter to master HEAD + xelis_dart_sdk 0.31.0

danrmiller:


danrmiller:
Superseded by #1424 (merged to staging as 78243c3f2).

Both pins this PR raised have since been taken further on staging:
- `xelis_dart_sdk` 0.30.9 -> 0.31.0 here; staging is now **0.35.1**. 0.31.0 was a cap imposed by `json_serializable` 6.11.2 pinning `json_annotation <4.10.0`, which was only in the graph because the cypherstack/tezart fork listed that codegen tool as a runtime dep. Fixed upstream in tezart as 84c56310, which lifted the cap without any codegen-toolchain realignment.
- `xelis_flutter` v0.2.1 -> `f67a02c` here; staging is now `3e5ac06c` (past f67a02c, and off the floating `branch = "dev"` Rust dep).

The `pubspec.lock` here has also gone stale and carries unrelated regressions (`cs_salvium_flutter_libs` and its platform packages downgraded 3.x -> 2.x, `coinlib`/`coinlib_flutter` and `flutter_libsparkmobile` pointing at refs staging does not use), so it is not mergeable as-is.

Nothing depends on the branch, so deleting it. The commit and diff stay accessible on this PR.

PR #1426, opened 2026-08-06
Fix intermittent "fee cannot be less than vSize" errors and Firo EX address sends

reubenyap:
Fixes intermittent fee-versus-vSize failures in legacy coin selection and restores Firo EX address sends through optimal selection.

## Changes

- Reconcile one-output, two-output, and send-all fees against the final signed vSize at the selected rate.
- Retry fixed-amount sends with another eligible UTXO when final signing increases the required fee.
- Return a built one-output transaction when the remaining value cannot form non-dust change.
- Clamp server estimates to each coin's existing default and update eCash's fallback to its [1000-atom/kB relay minimum](https://github.com/Bitcoin-ABC/bitcoin-abc/blob/677e37d9e325f9dd9d214eb0c7d0acbf49158ffe/src/policy/policy.h).
- Parse Firo EX addresses consistently for selection and transaction building.
- Keep caller-authoritative and MWEB fees unchanged.

Complementary to #1422, which fixes OP_RETURN sizing in optimal selection.

## Validation

- `flutter test --no-pub test/wallets/electrumx_interface_test.dart`: 4 passed.
- `flutter analyze --no-pub lib/wallets/crypto_currency/coins/ecash.dart lib/wallets/wallet/wallet_mixin_interfaces/electrumx_interface.dart test/wallets/electrumx_interface_test.dart`: no errors or warnings; 13 existing line-length infos.
- `dart format --output=none --set-exit-if-changed` on the three analyzed files: clean.
- `git diff refs/codex/pr1426-base HEAD --check`: clean.
- `flutter test --no-pub`: 344 passed, 4 skipped, and 1 unrelated existing failure in `create_pin_view_screen_test.dart`.

A native desktop build was not completed because CMake is unavailable in the local environment.

[GitHub Test run 33382300891](https://github.com/cypherstack/stack_wallet/actions/runs/33382300891) passed setup and changed-file formatting, then finished with 100 tests passed and 81 failed. The failures are compilation cascades because the workflow's fork fallback key stub omits four unrelated exchange constants. That workflow repair remains outside this PR.

### DIFF
--- a/lib/utilities/extensions/impl/big_int.dart
+++ b/lib/utilities/extensions/impl/big_int.dart
@@ -11,6 +11,8 @@
 import 'dart:typed_data';
 
 extension BigIntExtensions on BigInt {
+  BigInt atLeast(BigInt minimum) => this < minimum ? minimum : this;
+
   String get toHex {
     if (this < BigInt.zero) {
       throw Exception("BigInt value is negative");
--- a/lib/utilities/extensions/impl/big_int.dart
+++ b/lib/utilities/extensions/impl/big_int.dart
@@ -11,8 +11,6 @@
 import 'dart:typed_data';
 
 extension BigIntExtensions on BigInt {
-  BigInt atLeast(BigInt minimum) => this < minimum ? minimum : this;
-
   String get toHex {
     if (this < BigInt.zero) {
       throw Exception("BigInt value is negative");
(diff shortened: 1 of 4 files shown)


julian-CStack:
This should be handled with the fee calc changes in staging now

PR #1429, opened 2026-08-10
Epicbox Cancellation Requests & Multi-Party Reflection [DO NOT MERGE]

who-biz:
Do not merge this yet (mostly here for easily referenced diff and a working build tree). Much needs changed before production (epic repo dependency updates, epicbox server upgrades by all instance admins, etc). But all is functional and working well with the prospective upgrade codebases, so far in my testing. Will cover more testing tomorrow with a clean build too.  As I'm doing that, I'll write up a full-throated summary and replace this (placeholder) text.

I would also like to clean up wallet lifetime.  As it stands, wallet is instantiated with 
```rust
let wallet_int = Box::into_raw(Box::new(wlt)) as i64;
```
Then, no cleanup for the `Box` occurs even on `delete_wallet` calls. While I haven't yet tested to verify - seems like wallet lifetime remains for entirety of process execution, for every wallet scenario.  To this point: I'd also like to clean up `listener_cancel`. 

**Quick references for bundled submodules, dependencies, new epicbox server code:**
- **`flutter_libepiccash`**: https://github.com/who-biz/flutter_libepiccash/tree/ebox-cancel-tx
- **`epic-wallet`**: https://github.com/who-biz/epic-wallet/tree/ebox-cancel-txs
- **`epicboxnodejs`**: https://github.com/who-biz/epicboxnodejs/tree/cancel-tx

FFI was overhauled in these changes as well, for safer memory handling and full utilization of rust ownership for safety. Nullable pointers have been removed everywhere they could be in the FFI, and intermediate variables have been minimized as well. 

Local cancellation of slates is a fallback case now, and a persistent 32 character identifier is now chosen & assigned by sender to a given epicbox-relayed slatepack.  On `postSlate()` calls, the epicbox server persists this ID, alongside the existing `epicboxmsgid`s (message IDs change with each step of ITX). These are structurally different, intentionally.  That persistent ID becomes our cancellation mechanism for requests (signature-gated and requires proof of tx membership). 

Once a cancellation is requested, the epicbox server removes the slates-in-progress from slates DB, and adds a tombstone for them in a newly added `cancelledSlates` DB. Receiver or sender (if disconnected) is notified upon next connection to epicbox of the cancellation. Both receiver & sender can request a cancellation, and this is reflected to the other party over epicbox protocol extensions. The protocol message is wired into a local cancellation for both parties, which is gated by the newly added `epicbox_tx_id`. 

New methods involved for cancellation can be called from CLI as follows:
```sh
# by standard index, or slate, etc - via new epicbox flag on cancel cmd:
./epic-wallet cancel -i <tx_id> -m epicbox

# or by epicbox_tx_is
./epic-wallet cancel -e <epicbox_tx_id>
```


tool/wl_templates/EPIC_libepiccash_interface_impl.template.dart
```diff
@@ -124,6 +137,17 @@ final class _LibEpicCashInterfaceImpl extends LibEpicCashInterface {
       refreshFromNode: refreshFromNode,
     );
 
+    // Log the flutter_libepiccash Transaction BEFORE converting it.
```
who-biz, line 140: This should be removed before prod.

lib/utilities/address_utils.dart
```diff
@@ -153,20 +155,25 @@ class AddressUtils {
...
+      final Map<String, String> filteredParams = _filterParams(parsedData);
 
       return PaymentUriData(
         scheme: scheme,
-        address: parsedData['address']!.trim(),
-        amount: filteredParams['amount'] ?? filteredParams['tx_amount'],
-        label: filteredParams['label'] ?? filteredParams['recipient_name'],
-        message: filteredParams['message'] ?? filteredParams['tx_description'],
-        paymentId: filteredParams['tx_payment_id'],
-        // Specific to Monero
```
who-biz, line 169: 1.) check that the `parsedData['address']!.trim()` change to `address.trim()` doesn't cause issues for other coins.
2.) ensure comment `// Specific to Monero` is written back into PR code (we want diff to be absolutely minimal).

lib/utilities/address_utils.dart
```diff
@@ -153,20 +155,25 @@ class AddressUtils {
...
 
       // Normalize the URI scheme.
-      final String scheme = parsedData['scheme'] ?? '';
-      parsedData.remove('scheme');
+      final String scheme = parsedData["scheme"] ?? "";
+      parsedData.remove("scheme");
 
       // Filter out unrecognized parameters.
-      final filteredParams = _filterParams(parsedData);
+      final String? address = parsedData["address"];
```
who-biz, line 162: We do proper null-checking up here now - See comment on L169, item 1.

lib/wallets/wallet/impl/epiccash_wallet.dart
```diff
@@ -1386,10 +1389,22 @@ class EpiccashWallet extends Bip39Wallet {
       final slatesToCommits = info.epicData?.slatesToCommits ?? {};
 
       for (final tx in transactions) {
+
+        Logging.instance.w(
```
who-biz, line 1393: This needs removed before prod.

PR #1430, opened 2026-08-13
flatpak: add :create to data-dir filesystem grants so first-run persists

danrmiller:


danrmiller:
This is just for the CI builds, the release process generates its own manifests.

PR #1434, opened 2026-08-25
build: re-pin devicelocale to a reachable upstream commit

sneurlax:
ba7d7d87 no longer exists in cypherstack/flutter-devicelocale, so pub cannot resolve it from a cold cache and CI fails at "Get dependencies" before running anything. 73c4ed94 is the linux-changes head, which already merges master and keeps the Devicelocale.currentAsLocale API the app uses.

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -801,11 +801,11 @@ packages:
     dependency: "direct main"
     description:
       path: "."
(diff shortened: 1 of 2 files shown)


julian-CStack:
linux fix was merged upstream so use pub dev hosted version instead.
https://github.com/cypherstack/stack_wallet/pull/1435

PR #1442, opened 2026-09-15
Add SafeArea for navigation bar user

ErgoTester:


julian-CStack:
Thanks! This is the right direction but there are a couple little issues:

1. One of the safe areas wraps the entire sheet child leaving a space below the sheet on some devices.
2. The transaction fees sheet is shown in other places that this specific change doesn't fix

Rather than deal with merge conflicts later I've moved the safe area wrappers on my current working branch: https://github.com/cypherstack/stack_wallet/commit/04694bc54eb26febcda5b70dd45cd3aaf781bc3f

PR #1444, opened 2026-09-16
feat: add rsFIRO Ethereum token

reubenyap:
## Summary

- add rsFIRO to the default Ethereum token list using the official contract address and 8 decimals
- let Campfire users add Firo and Ethereum wallets
- limit Campfire Ethereum token choices to rsFIRO, USDC, and USDT while keeping Stack Wallet behavior unchanged
- add missing default tokens to existing local token databases

Contract metadata: https://firo.org/2026/09/15/rsfiro-live.html

## Testing

- GitHub Actions test job passes
- Dart formatting check passes
- default-token allowlist test covers the Campfire Ethereum tokens

CI: https://github.com/cypherstack/stack_wallet/actions/runs/35188999250/job/105097180090


julian-CStack:
We should avoid any more `AppConfig.appName == "Campfire"` checks when possible. The existing checks were absolutely required to be able to migrate correctly from the old Campfire. It would be better to make use of the appconfig stuff for this where we can have default token contracts lists as well as options to have/not have add/remove tokens. 

The PR message mentions the contracts list check and mutation functionality does not change. This is false. Previously if a user removed a number of contract addresses while leaving at least one contract address then the code would leave that unchanged and with these changes it reinserts all default addresses back every time.

Closing this as it would be better to start fresh.

PR #1446, opened 2026-09-17
ci: guard every branch that carries work, and write down the layout

bitfinitechain:
First merge through the new arrangement, so the arrangement gets tested by using it.

Moves both workflows onto the branches that now carry work. They fired only on staging, which under the new layout is the one branch nothing lands on.

Adds docs/branching.md: what each branch is for, why the default moved off staging, and the three conditions an upgrade must meet to install over an existing wallet without disturbing it.

main requires the test job to pass before anything merges, so this PR is also the check that the protection works.


bitfinitechain:
Opened by mistake against the wrong repository. This belongs on our own fork, not upstream. Apologies for the noise.

